PR #311, opened 2020-05-06
No longer ignore output next time if ncycle is zero

### LINKED ISSUE #310: Output next time is ignored on startup
**Prerequisite checklist**

Place an X in between the brackets on each line as you complete these checks:
- [x] Did you check that the issue hasn't already been reported?
- [x] Did you check the documentation in the Wiki for an answer?
- [x] Are you running the latest version of Athena++?

**Summary of issue**

The if statement here purposely ignores  next time if `pm->time == pm->start_time`:
https://github.com/PrincetonUniversity/athena/blob/195f18321f530f21cba65afc99a4081426023850/src/outputs/outputs.cpp#L723
This forces the output and increments next time.

This is annoying if you want an output to start at a particular time.

**Steps to reproduce**

Use any pgen and set an output `next_time` > 0.

**Expected outcome**

First output is produced at `next_time`.

**Actual outcome**

First output is produced at t=0/start time. Second output is produced at `next_time` + `dt`. No output is actually produced at `next_time`.

**Possible solutions**
1) Remove `pm->time == pm->start_time`
2) Add a flag to not output at start time
3) Still output at `start_time` but don't increment `next_time` if it's greater than `start_time` (personally disfavored)

msbc:
## Prerequisite checklist

## Description
If `ncycle` is zero, output `next_time` is not ignored.

Closes #310

### DIFF
--- a/src/outputs/outputs.cpp
+++ b/src/outputs/outputs.cpp
@@ -720,7 +720,7 @@ void Outputs::MakeOutputs(Mesh *pm, ParameterInput *pin, bool wtflag) {
   bool first=true;
   OutputType* ptype = pfirst_type_;
   while (ptype != nullptr) {
-    if ((pm->time == pm->start_time) ||
+    if ((pm->time == pm->start_time && pm->ncycle) ||
         (pm->time >= ptype->output_params.next_time) ||
         (pm->time >= pm->tlim) ||
         (wtflag && ptype->output_params.file_type == "rst")) {


felker:
@tomidakn what is the role of `wtflag` in:

https://github.com/PrincetonUniversity/athena/blob/91154cd92f8d8dd171f2cc4eaf0ece126b3152d7/src/outputs/outputs.cpp#L774-L795

and what does it stand for? Maybe we should rename it. 

Seems to be only passed as `true` in `main.cpp` for the final output, but I don't see it being used in any of the output types?

kahoooo:
@felker Some quick guess: wall time flag? And wtflag default to be false in outputs.hpp?

tomidakn:
@kahoooo  is correct.

PR #363, opened 2021-04-26
Correct calculations of the energy viscous flux

tomo-ono:
When the orbital advection (OA) is enabled, we use the orbital system. However, the viscous flux is calculated from the primitives in the regular system rather than the orbital system. There were bugs to calculate the energy viscous flux because some parts of the calculations need the velocity field in the orbital system. This PR fix the bugs. This PR also includes a minor bug in `orbital_system_conversion.cpp`.


buildbot-princeton:

PR #378, opened 2021-07-13
Fix shearing boundary flag for corner neighbors

tomo-ono:
According to #337, the shear boundary flag was not set properly for corner neighbors. This PR fix the bug. Even using this PR, neither the simulation results nor the limitation of the shear periodic boundary with SMR/AMR has any change.

### DIFF
--- a/src/bvals/bvals_base.cpp
+++ b/src/bvals/bvals_base.cpp
@@ -713,11 +713,19 @@ void BoundaryBase::SearchAndSetNeighbors(MeshBlockTree &tree, int *ranklist,
         int nlevel = neibt->loc_.level;
         nblevel[l+1][m+1][n+1] = nlevel;
         if (nlevel >= loc.level || (myox1 == n && myox2 == m && myox3 == l)) {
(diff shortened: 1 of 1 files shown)


src/bvals/bvals_base.cpp
```diff
@@ -713,11 +713,19 @@ void BoundaryBase::SearchAndSetNeighbors(MeshBlockTree &tree, int *ranklist,
         int nlevel = neibt->loc_.level;
         nblevel[l+1][m+1][n+1] = nlevel;
         if (nlevel >= loc.level || (myox1 == n && myox2 == m && myox3 == l)) {
+          bool shear = false;
+          if ((nlevel == loc.level) &&
+              ((n == -1 && block_bcs[BoundaryFace::inner_x1]
+                                 == BoundaryFlag::shear_periodic) &&
```
changgoo, line 719: I think this should be `||`

changgoo: I mean `&&` at the end of L719

PR #380, opened 2021-08-19
Low-dissipation HLLC/HLLD solvers (Minoshima et al. 2021)

tomidakn:
These are newly proposed low-dissipation and Carbuncle-free extension of the HLLC (for hydro) and HLLD (for MHD) approximate Riemann solvers. The additional cost is only a few %.

Reference: Minoshima et al. 2021, arXiv:2108:04991 (JCoPh accepted)

## Prerequisite checklist
- [x] I have added tests to cover my changes. (quirk.cpp)

## Description
Besides the new LHLLC/LHLLD solvers, I moved `OutputCycleDiagnostics` to be called before `UserWorkAfterLoop` in the finalizing process in main.cpp.

## Testing and validation
quirk.cpp is the Quirk test (Quirk 1994), in which a strong shock with odd-even perturbation propagates aligned to the grid. This odd-even perturbation should not grow physically, but standard high-resolution schemes such as HLLC/HLLD/Roe fail. The new LHLLC/LHLLD schemes are free from this issue. This test code checks the growth of the perturbation.
...

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -98,7 +98,7 @@ parser.add_argument('--eos',
 # --flux=[name] argument
 parser.add_argument('--flux',
                     default='default',
-                    choices=['default', 'hlle', 'hllc', 'hlld', 'roe', 'llf'],
+                    choices=['default', 'hlle', 'hllc', 'lhllc', 'hlld', 'lhlld', 'roe', 'llf'], # noqa
                     help='select Riemann solver')
 
 # --nghost=[value] argument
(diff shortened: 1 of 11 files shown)


felker (approved):
All these changes look good; I look forward to reading the paper. 

Is there any reason why we shouldnt make these the default Riemann solvers? Are they fully compatible with SR and GR via frame transformations?

The following Wiki pages should probably be updated:
- https://github.com/PrincetonUniversity/athena/wiki/Configuring
- https://github.com/PrincetonUniversity/athena/wiki/1D-Shock-Tube
- https://github.com/PrincetonUniversity/athena/wiki/Special-Relativity
- https://github.com/PrincetonUniversity/athena/wiki/General-Relativity

c-white:
The new algorithm is strictly Newtonian, so can't be directly applied to SR/GR. Perhaps someone should write that paper, though. It took 4 years to go from HLLD to relativistic HLLD; maybe this time the wait can be shorter.

Actually, it's not clear how often this is needed in relativity anyway. Relativists rarely see carbuncle, since we have enough problems on our hands without looking for more ;) Also incompressible flows don't exist in relativity, and it seems these new solvers are most helpful in that limit.

tomidakn:
These solvers actually contain two independent improvements - one is Carbuncle care and the other is improvement in the highly subsonic limit.

I hesitate to make them default at this moment, as they are very new and not very well tested in general situations. I actually saw a case in which these solvers suppress physical instability in a shock with strong cooling. In such as case, as the shock is physically unstable, the Carbuncle does not persist so we do not need the care. We may consider making the subsonic improvement default, but this correction actually loses the good nature of the HLLC/D solver that all the intermediate states satisfy physical jump conditions, although the drawback from this violation seems negligibly small in most cases.

And as Chris says, these are not relativistic.

tomidakn:
I added brief description in https://github.com/PrincetonUniversity/athena/wiki/Configuring .

jmstone (approved):
All looks good to me.  I am also hesitant to make these the default solvers.  Is the fix for the carbuncle instability in these solvers better than the H-correction in Athena?  Your comment that is can sometimes suppress physical instabilities is worrisome.  I think it is great they are added, and I think people should try them out for applications.  Once we have some experience with them we could consider making them the defaults.

Overall the pull request looks great and you should go ahead and merge!

felker:
@tomidakn a follow-up suggestion: could we add a Python regression test based on `quirk.cpp`?

tomidakn:
@jmstone  I think this is better than the original H-correction, because the H-correction add numerical viscosity at everywhere near velocity discontinuities, but the new one adds viscosity only where the shock is aligned to the grid. Also, the new solver only corrects the sound speed in the intermediate state (S_M), it retains the overall structure of the original HLLD. So I think this one is less invasive. 

@felker  To be honest I am not very good at Python. Can you do it for me? quirk.cpp tells you if the selected scheme passed or failed the test. In hydro, Roe and HLLC should fail, while the others should pass. For MHD, HLLD and Roe should fail.

felker:
@tomidakn ok I will quickly throw something together. LLF is also stable. 

The test can only be run with a single MeshBlock, right?

Issue #393, opened 2021-09-21
plot_slice.py issues


kurtsansom:
**Prerequisite checklist**

Place an X in between the brackets on each line as you complete these checks:
- [X] Did you check that the issue hasn't already been reported?
- [X] Did you check the documentation in the Wiki for an answer?
- [X] Are you running the latest version of Athena++?

**Summary of issue**
athdf reader appears to read as byte objects rather than strings objects.

when reading a file in cylindrical coordinates It returns a warning, `warnings.warn('Coordinates not recognized; results may be misleading')` but when I print the string it reveals the string read from the file is a byte string.
 e.g. b'cylindrical' which when compared with 'cylindrical' returns false. 

**Steps to reproduce**

Configuration:

```
configure = --prob=cme --prob=cme --eos=adiabatic --coord=cylindrical -hdf5 -omp -b --hdf5_path=/home/krs/miniconda3/envs/mhd
```

```
../athena/vis/python/plot_slice.py -d 3 --colormap plasma --stream Bcc  solar_wind.out3.00001.athdf dens output/slice.png
```

**Actual outcome**

**Additional comments and/or proposed solution**

**Version info**
git tag: 5261e49

Conda environment:
[conda_env.txt](https://github.com/PrincetonUniversity/athena/files/7204690/conda_env.txt)

c-white:
This seems like one of those things Python3 incomprehensibly broke. Apparently either all other users are like me and still use Python2, or no one else uses the plotting scripts...

Can you checkout the branch `plotting_fix` to make sure it fixes the problem?

felker:
Actually, I am pretty sure this is due to changes in `h5py` with the release of their 3.x versions:
- https://docs.h5py.org/en/latest/strings.html
- https://github.com/tensorflow/tensorflow/issues/44467

msbc:
I feel like we should include these plotting scripts in regression testing. 

Also, to add onto @felker 's point: I'm pretty sure I successfully used `plot_slice.py` in the past with Python3. While I don't really use the scripts in `vis/python`, I've refused to use Python2 for several years now and I'd be surprised if I hadn't used `plot_slice.py` in that time.

kurtsansom:
@c-white  why not use `utf-8`?

kurtsansom:
@c-white the branch works for me when doing the basic plot. 

[...]

PR #394, opened 2021-09-21
Update plotting slices and add polar projection view 

kurtsansom:
## Prerequisite checklist

- [ ] My code follows the Athena++ [Style Guide](https://github.com/PrincetonUniversity/athena/wiki/Style-Guide)

## Description
- Newer versions of h5py read strings as by strings and require being encoded to enable/disable options, prefer utf-8 over ascii
- Stream plots from single precision data fail without recasting grid limits, so recast cell coordinates into uniform data
- prefer explicit objects over `plt.` usage with figures, axes, etc. 
- plots assume rectangular grid, add polar projection for cylindrical plots
- expose num_ghosts as a parameter
- resolve https://github.com/PrincetonUniversity/athena/issues/393

Commits should have been split up but want to get feedback to determine if these changes are useful upstream

## Testing and validation
Ad hoc testing on current data, appears to work for cylindrical slice 

## To-do

- [ ] currently the figure size is hard-coded to `(12, 12)`  and could be made a variable with a default

### DIFF
--- a/vis/python/athena_read.py
+++ b/vis/python/athena_read.py
@@ -216,7 +216,7 @@ def vtk(filename):
     # Read raw data
     with open(filename, 'rb') as data_file:
         raw_data = data_file.read()
-    raw_data_ascii = raw_data.decode('utf-8', 'replace')
+    raw_data_ascii = raw_data.decode('ascii', 'replace')
 
     # Skip header
     current_index = 0
@@ -365,10 +365,10 @@ def athdf(filename, raw=False, data=None, quantities=None, dtype=None, level=Non
             data['x3v'] = f['x3v'][:]
 
             # Get metadata describing file layout
-            dataset_names = np.array([x.decode('utf-8', 'replace')
+            dataset_names = np.array([x.decode('ascii', 'replace')
                                       for x in f.attrs['DatasetNames'][:]])
             dataset_sizes = f.attrs['NumVariables'][:]
-            variable_names = np.array([x.decode('utf-8', 'replace')
+            variable_names = np.array([x.decode('ascii', 'replace')
                                        for x in f.attrs['VariableNames'][:]])
 
             # Store cell data
@@ -445,7 +445,7 @@ def athdf(filename, raw=False, data=None, quantities=None, dtype=None, level=Non
                 num_extended_dims += 1
 
         # Set volume function for preset coordinates if needed
-        coord = f.attrs['Coordinates'].decode('utf-8', 'replace')
+        coord = f.attrs['Coordinates'].decode('ascii', 'replace')
         if level < max_level and not subsample and not fast_restrict and vol_func is None:
             x1_rat = f.attrs['RootGridX1'][2]
             x2_rat = f.attrs['RootGridX2'][2]
--- a/vis/python/athena_read.py
+++ b/vis/python/athena_read.py
@@ -216,7 +216,7 @@ def vtk(filename):
     # Read raw data
     with open(filename, 'rb') as data_file:
         raw_data = data_file.read()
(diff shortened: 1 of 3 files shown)


c-white:
I think these are generally useful.

Quickly scanning them, I think ascii -> utf-8 is probably fine, though I'm pretty certain Athena will never use non-ascii characters when writing that string to file. @felker, since you put that in originally, is there a preference? In any event, this should be done consistently with decodings in `athena_read.py`.

c-white:
One other thing to consider: Having a polar plot is definitely useful, but there should probably be a way for users to turn it off as well. It's sometimes helpful to see the data in its raw form without manipulating it into a more physical shape, and if `plot_slice.py` doesn't do that, no other script will.

Alternatively, we could do something like spherical coordinates does: `plot_slice.py` just plots raw slices, while a new `plot_cylindrical.py` makes more physically meaningful plots. This might be overkill for cylindrical coordinates, though.

vis/python/athena_read.py
```diff
@@ -216,7 +216,7 @@ def vtk(filename):
     # Read raw data
     with open(filename, 'rb') as data_file:
         raw_data = data_file.read()
-    raw_data_ascii = raw_data.decode('ascii', 'replace')
```
felker, line 219: VTK binaries are ASCII encoded, not UTF-8.

felker (changes requested):
I agree with using `fig`, `ax`, etc. instead of `plt` for the matplotlib calls.

The decode calls should be to `ascii`, not `utf-8`.

https://support.hdfgroup.org/HDF5/doc/RM/RM_H5T.html#Datatype-SetCset
> By default, HDF5 creates object and attribute names with ASCII character encoding. An object or attribute creation property list setting is required to create object names with UTF-8 characters. This uses the function H5Pset_char_encoding, which sets the character encoding used for object and attribute names.

We do not specify UTF in the 3x string attributes written in https://github.com/PrincetonUniversity/athena/blob/master/src/outputs/athena_hdf5.cpp, so they are ASCII encoded. See, e.g.:
```
➜  athena git:(master) h5dump -A  kh-shear-lecoanet.out2.00027.athdf
HDF5 "kh-shear-lecoanet.out2.00027.athdf" {
GROUP "/" {
   ATTRIBUTE "Coordinates" {
      DATATYPE  H5T_STRING {
         STRSIZE 21;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_ASCII;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SCALAR
      DATA {
      (0): "cartesian"
      }
   }
   ATTRIBUTE "DatasetNames" {
      DATATYPE  H5T_STRING {
         STRSIZE 21;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_ASCII;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SIMPLE { ( 1 ) / ( 1 ) }
      DATA {
      (0): "prim"
      }
...
   ATTRIBUTE "VariableNames" {
      DATATYPE  H5T_STRING {
         STRSIZE 21;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_ASCII;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SIMPLE { ( 6 ) / ( 6 ) }
      DATA {
      (0): "rho", "press", "vel1", "vel2", "vel3", "r0"
      }
   }
```

felker:
@kurtsansom did you see my feedback on this? Can we update and merge soon?

kurtsansom:
@felker I am behind on some other projects at the moment. I can close for now or someone else can pick it up.

PR #412, opened 2021-12-14
Extend gravity to allow enrolling multiple source fields for gravitational potential calculation

kahoooo:
## Prerequisite checklist

## Description

In some cases, one may want to have extra sources of gravity on top of hydro's self-gravity.
Some examples are sink particles' mass being deposit onto the mesh, dust particles' self-gravity (or maybe dark matter?).
A separate potential calculation for each physics would be expensive and a simple modification like this pull request can combine all calculation into one by (un-)enrolling sources between potential calculation.

Hydro density field is kept enrolled by default. Other density field can be enrolled using
`pgrav->EnrollSource(arr, idx)`
and density fields can be unenrolled using
`pgrav->UnenrollSource(arr, idx)`

Density field can also be enrolled in a subset of all meshblocks to reduce overhead.
In case only one field is enrolled, the accumulated source field would simply reference the only enrolled field so no overhead is expected.

1. Add 3 class functions in gravity class (enroll, unenroll, accumulate)
2. Allocate buffer for storing accumulated density field
3. Minor changes in meshblock.cpp to enroll hydro's density by default
4. Minor changes in outputs.cpp to allow outputting accumulated field

## Testing and validation

...

## To-do

- [ ] Future development task

### DIFF
--- a/src/gravity/gravity.hpp
+++ b/src/gravity/gravity.hpp
@@ -50,8 +50,6 @@ class Gravity {
   void Initialize(ParameterInput *pin);
   void Solver(const AthenaArray<Real> &u);
 
-  void EnrollSource(AthenaArray<Real> &arr, int idx);
-  void UnenrollSource(AthenaArray<Real> &arr, int idx);
   void ComputeSource();
 
   friend class MGGravityDriver;
@@ -60,9 +58,8 @@ class Gravity {
   bool gravity_tensor_momentum_;
   bool gravity_tensor_energy_;
 
-  // actual copy of accumulated source and (pointers, index) pair to enrolled sources
+  // actual copy of accumulated source
   AthenaArray<Real> accumulated_src_;
-  std::vector<std::pair<AthenaArray<Real>*, int>> enrolled_src_;
 };
 
 #endif // GRAVITY_GRAVITY_HPP_
--- a/src/gravity/gravity.hpp
+++ b/src/gravity/gravity.hpp
@@ -12,6 +12,8 @@
 // C headers
 
 // C++ headers
+#include <utility>
+#include <vector>
 
 // Athena++ headers
 #include "../athena.hpp"
@@ -34,6 +36,7 @@ class Gravity {
   Gravity(MeshBlock *pmb, ParameterInput *pin);
 
   MeshBlock* pmy_block;  // ptr to MeshBlock containing this Field
+  AthenaArray<Real> src;   // shallow copy of accumulated source
   AthenaArray<Real> phi;   // gravitational potential
   AthenaArray<Real> empty_flux[3];
   Real gconst, four_pi_G;
@@ -46,11 +49,19 @@ class Gravity {
   void Initialize(ParameterInput *pin);
   void Solver(const AthenaArray<Real> &u);
 
+  void EnrollSource(AthenaArray<Real> &arr, int idx);
+  void UnenrollSource(AthenaArray<Real> &arr, int idx);
+  void ComputeSource();
+
   friend class MGGravityDriver;
 
  private:
   bool gravity_tensor_momentum_;
   bool gravity_tensor_energy_;
+
+  // actual copy of accumulated source and (pointers, index) pair to enrolled sources
+  AthenaArray<Real> accumulated_src_;
+  std::vector<std::pair<AthenaArray<Real>*, int>> enrolled_src_;
 };
 
 #endif // GRAVITY_GRAVITY_HPP_
--- a/src/gravity/gravity.hpp
+++ b/src/gravity/gravity.hpp
@@ -41,6 +41,7 @@ class Gravity {
   AthenaArray<Real> empty_flux[3];
   Real gconst, four_pi_G;
   bool srcterm;
+  bool output_source;
 
   // TODO(felker): consider creating a CellCentered.. derived class, and changing to
   //GravityBoundaryVariable *pgbval;
--- a/src/mesh/meshblock.cpp
+++ b/src/mesh/meshblock.cpp
@@ -155,7 +155,6 @@ MeshBlock::MeshBlock(int igid, int ilid, LogicalLocation iloc, RegionSize input_
   if (SELF_GRAVITY_ENABLED) {
     // if (this->grav_block)
     pgrav = new Gravity(this, pin);
-    pgrav->EnrollSource(phydro->u, IDN);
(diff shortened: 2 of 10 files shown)


felker:
Could you add an example problem that utilizes this functionality? We would need to document it in the Wiki, too.

src/gravity/gravity.hpp
```diff
@@ -46,11 +50,19 @@ class Gravity {
...
+
   friend class MGGravityDriver;
 
  private:
   bool gravity_tensor_momentum_;
   bool gravity_tensor_energy_;
+
+  // actual copy of accumulated source and (pointers, index) pair to enrolled sources
+  AthenaArray<Real> accumulated_src_;
+  std::vector<std::pair<AthenaArray<Real>*, int>> enrolled_src_;
```
changgoo, line 65: I see this way of pointing the source array confusing. Isn't it better to point actual source array (e.g., density) rather than pointing 4D array with index?

src/mesh/meshblock.cpp
```diff
@@ -155,6 +155,7 @@ MeshBlock::MeshBlock(int igid, int ilid, LogicalLocation iloc, RegionSize input_
   if (SELF_GRAVITY_ENABLED) {
     // if (this->grav_block)
     pgrav = new Gravity(this, pin);
+    pgrav->EnrollSource(phydro->u, IDN);
```
changgoo, line 158: If you agree with the comment on the style of `enrolled_src_`, this can be enrolling `den` as a shallow slice of `u`.

changgoo (commented):
I see this new functionality provides nice interface to add sources for the gravity solver. I feel that pairing 4D array with index is a bit complicated. This can be more straightforward if it assumes that the user will take correct shallow slice to enroll (with clear documentation and example for default gas density).

kahoooo:
> Could you add an example problem that utilizes this functionality? We would need to document it in the Wiki, too.

Sure! Will do once we agree on how this should be implemented.


> I see this new functionality provides nice interface to add sources for the gravity solver. I feel that pairing 4D array with index is a bit complicated. This can be more straightforward if it assumes that the user will take correct shallow slice to enroll (with clear documentation and example for default gas density).

After some thinking, I think what it implies is that the `Hydro` class would need to keep a non-destructed shallow copy of the density.
Also, during integration, since `SwapAthenaArray()` is used on the 4D actual copy, the 3D shallow copy may not be pointing to the buffer correctly unless `Hydro` manage it.

So the change may be quite intrusive. What do you think?

changgoo:
Ah, okay. If that's the case, the current design would be better. I've added this feature to our current developing fork with particles and passed all regression tests. Not sure whether @tomidakn has any opinion regarding its relation to AMR multigrid solver. Otherwise, it looks good to me.

tomidakn:
I'm sorry for my slow response.

First, I agree this is a nice feature to have, and it is nicely encapsulated. And this should work with AMR without any problem. 

On the other hand, I am a bit concerned about the performance. In this implementation, a user has to prepare arrays to enroll extra sources. This costs some memory and operations. Instead, if we allow direct accumulation of extra sources into the source array, we may save this cost. I was thinking of something like that for Particle Mesh gravity. How do you think?

kahoooo:
There are indeed some tradeoff in this implementation and I am open to other implementation if we prefer that.
Currently, I can only think of these three following designs:
1. Current implementation: each enrolled source is in its own array -> `gravity` responsible for the accumulation
2. Enroll custom functions to modify the source array directly (including `hydro`)
3. Combination of both: allow both arrays and functions to be enrolled

Other than (1) being the most memory-hungry, the only other drawback is that (2) would perform worse if only `hydro` is enrolled simply because (2) requires addition which is not needed in (1) and (3) (slicing is done instead).

I don't know... I would say if performance is a concern, we may want (3) I guess?

pdmullen (commented):
I agree with everything that was posted elsewhere.  I just wanted to point out the potentially obvious: adding an additional source field for the gravitational potential will violate total momentum conservation.

tomidakn:
Well, either (1) or (2) is general enough, I do not think we need both. I prefer (2) , but not by far. If we load the hydro density by default and call this user-defined source manipulation function only when it is defined, there should be almost no performance loss, I guess. But I'm open to other ideas.

> Patrick, you are right, but  I would say it's the user's responsibility.

kahoooo:
The only concern that I have is that if, by loading hydro density, it means having the source array initialized as the hydro density, it will have additional copy operations compared with the implementation before this pull request (although it would be minimal).

I suggest having a similar implementation where the source array is initialized as a shallow array if not user defined function is enrolled. Otherwise, copy hydro density to source and call user-defined function to manipulate the source array.

I will draft it and push the commits here tonight.

tomidakn:
Ah, I see, I guess I was missing your point. In the current implementation, the source term is initialized with deep copy. While this is not necessary at this moment, I implemented this intentionally for future extension. For example, technically we can overlap the gravity solver and MHD flux calculation using different cores.  In this case it is better to allocate a separate array for the source to make them independent. Also, I anticipated that a particle mesh solver needs to add the particle density to the source, which should be kept separately from the hydro density. So, I would like to keep this structure, and then my preference is Option (2).

kahoooo:
Now I see the point. In that case it does make more sense implementing option (2).
But just making sure, when referring to the source array, are you referring to that currently owned by multigrid solver and FFT solver or that by gravity as implemented in this pull request?

tomidakn:
I meant the internal source array (pmg->src) in Multigrid objects, as originally Gravity object did not have the accumulated_src_ array.

Um, I actually had to add one more point - the number of ghost cells in Multigrid is different from NGHOST. So for MG we need deep copy anyway at some point.

I guess the simplest implementation is to modify pfft->LoadSource and pmg->LoadSource to manipulate the source.

kahoooo:
Sorry for the slow response. It took me a while to think through it.
I wasn't able to come up with a way that doesn't require an additional deep copy: the way that pfft->LoadSource and pmg->LoadSource initialize their source arrays and the source arrays themselves are vastly different too different. I doubt it is feasible to have one single function manipulating both arrays correctly.

So I have this version that still have some tradeoff.
Here, `gravity` keep a copy of source array which requires one extra deep copy when there are functions enrolled. This should have less memory footprint and require fewer operations than the previous implementation of this pull request.

kahoooo:
Closing until better implementation is figured out...

PR #413, opened 2021-12-14
Allow cherrypicking variables to be output

kahoooo:
## Prerequisite checklist

## Description

Sometimes one would like to cherrypick variables to be output to reduce output sizes when a lot of variables are output by default. But outputting variables into multiple files generates unmanageable amount of files (plus there will be redundant information such as the mesh coordinates).

This pull request allows users to provide a list of variables in a form of comma-delimited string in athinput files, e.g.,
```
variable = d,p,E,m,bcc
```

It changes `OutputType` to look for occurrences instead of exact matches.

## Testing and validation

...

## To-do

- [ ] Future development task

### DIFF
--- a/src/outputs/outputs.cpp
+++ b/src/outputs/outputs.cpp
@@ -1069,3 +1069,15 @@ void OutputType::CalculateCartesianVector(AthenaArray<Real> &src, AthenaArray<Re
   }
   return;
 }
+
+bool OutputType::ContainVariable(const std::string &haystack, const std::string &needle) {
+  if (haystack.compare(needle) == 0)
+    return true;
+  if (haystack.find(',' + needle + ',') != std::string::npos)
+    return true;
+  if (haystack.find(needle + ',') == 0)
+    return true;
+  if (haystack.find(',' + needle) == haystack.length() - needle.length())
+    return true;
+  return false;
+}
--- a/src/outputs/outputs.cpp
+++ b/src/outputs/outputs.cpp
@@ -1077,7 +1077,7 @@ bool OutputType::ContainVariable(const std::string &haystack, const std::string 
     return true;
   if (haystack.find(needle + ',') == 0)
     return true;
-  if (haystack.find(',' + needle) == haystack.length() - needle.length())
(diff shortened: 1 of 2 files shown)


src/outputs/outputs.cpp
```diff
@@ -1069,3 +1069,15 @@ void OutputType::CalculateCartesianVector(AthenaArray<Real> &src, AthenaArray<Re
...
 }
+
+bool OutputType::ContainVariable(const std::string &haystack, const std::string &needle) {
+  if (haystack.compare(needle) == 0)
+    return true;
+  if (haystack.find(',' + needle + ',') != std::string::npos)
+    return true;
+  if (haystack.find(needle + ',') == 0)
+    return true;
+  if (haystack.find(',' + needle) == haystack.length() - needle.length())
```
felker, line 1080: This ensure that `...,b` is not considered a match for `...,b3`, right?

felker (approved):
This is helpful, thanks!

felker:
If you can provide me edits for https://github.com/PrincetonUniversity/athena/wiki/Outputs (if you are unable to edit it directly yourself), I will apply them and we can merge this.

c-white:
This would also warrant small changes to https://github.com/PrincetonUniversity/athena/wiki/Outputs and https://github.com/PrincetonUniversity/athena/wiki/HDF5-Format

src/outputs/outputs.cpp
kahoooo: Yes, that's the intention... although just found that the index was off by one.
Just fixed it in the last commit.

kahoooo:
Let me also try to make it compatible with user_out_var...

kahoooo:
Just tested it quickly with linear_wave and it seems to be working properly.
Also I can't find the option to edit the wiki page. But this change under "Output Variables" should suffice:

> Output Variables
> A single variable, a set (`cons` vs. `prim` vs. ...) of variables, or a combination of both can be specified in the variable field. To select multiple variables, a list of variable separated by a single comma `,` should be provided (e.g., to select conservative density and pressure, use `D,p`). Note that the values below are case-sensitive:

PR #417, opened 2021-12-23
Added "dcycle" option to output data at constant cycle intervals

tomidakn:
This PR adds "dcycle" option in the <output> block in input. "dcycle" is an integer number, and the code outputs data at every "dcycle" step. "dt" and "dcycle" are exclusive.

## Prerequisite checklist

## Description
This feature is useful for some problems with drastic change in time scale. A particular example is self-gravitational collapse like star formation.

### DIFF
--- a/src/outputs/outputs.cpp
+++ b/src/outputs/outputs.cpp
@@ -135,9 +135,22 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin) {
 
       // set time of last output, time between outputs
       op.next_time = pin->GetOrAddReal(op.block_name,"next_time", pm->time);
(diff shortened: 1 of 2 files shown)


jmstone (approved):
I like this added feature, and in fact I have often wished we had it when using the code myself in the past.  I approve!  And I will add this feature to the Kokkos version.

felker (approved):
LGTM. 

Small detail, but I am not sure of what happens if the user specifies a negative `dcycle`? Negative output `dt` is fine as a quick way to disable a specific output format. 

Is the Wiki updated?

tomidakn:
I updated the document.

The output is skipped if dcycle is negative. Same as negative dt.

PR #425, opened 2022-01-26
Tiny fix for reading Schwarzschild .athdf files

c-white:
`athena_read.py` was missing a pair of parentheses, resulting in erroneous values for `x1v` being returned. This only affected Schwarzschild coordinates, which are rarely used in production.


felker:
Ready to merge?

c-white:
This can be merged at any time.

PR #434, opened 2022-04-10
fix dependency of passive scalar boundary communication

tomo-ono:
In AddSourceTermsHydro, it is supposed to impose source functions not only on hydro variables but also on passive scalars. If we change passive scalars in AddSourceTermsHydro, SEND_SCLR and SETB_SCLR need to wait for the execution of SRCTERM_HYD rather than INT_SCLR. This PR fixes the dependencies. 
The name of AddSourceTermsHydro is confusing. I think that we should make a new function as AddSourceTermsScalars or rename the function.

### DIFF
--- a/src/task_list/time_integrator.cpp
+++ b/src/task_list/time_integrator.cpp
@@ -972,9 +972,9 @@ TimeIntegratorTaskList::TimeIntegratorTaskList(ParameterInput *pin, Mesh *pm) {
         AddTask(RECV_SCLR,NONE);
         AddTask(SETB_SCLR,(RECV_SCLR|CALC_HYDORB));
       } else {
-        AddTask(SEND_SCLR,INT_SCLR);
+        AddTask(SEND_SCLR,SRCTERM_HYD);
         AddTask(RECV_SCLR,NONE);
-        AddTask(SETB_SCLR,(RECV_SCLR|INT_SCLR));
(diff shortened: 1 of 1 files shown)


felker:
I agree with the proposed changes, maybe rename to `AddSourceTermsHydroScalars` for now (since this function is not called by users anyway), consider adding a new function `AddSourceTermsScalars` and task `SRCTERM_SCLR` in the future? Thoughts @tomidakn @pdmullen @c-white ?

The task list construction is very hard to read now, but this seems like the correct fix in analogy to: 
https://github.com/PrincetonUniversity/athena/blob/49d491e1837a578ebf29216c7cdade8e842cb9fc/src/task_list/time_integrator.cpp#L945-L947

c-white:
I can imagine cases where the hydro variables and passive scalars have tightly coupled source terms, for instance if the cooling rate depends on composition. So splitting the function *might* make some use cases more awkward.

felker:
ah yes, forgot about that case @c-white. Let's just rename the function and task ID for now. @tomo-ono want to open a new PR with those changes?

tomo-ono:
OK. I will do it. Does anyone have a good idea for the function name?

felker:
`AddSourceTermsHydroScalars` and `SRC_TERMS`?

buildbot-princeton:

PR #440, opened 2022-04-24
Fix bug specifying uov output

### LINKED ISSUE #436: Small bug in specifying uov output
Hello,
It seems after commit 33ad4553d023ffbf1006aba6ec67c8d68e0b8d51, just specifying uov as an output variable no longer works (throws a variable not implemented error), which probably wasn't the intended effect since the wiki still shows this as an example. Instead one is forced to cherry pick the user output variables (so uov0 works). This is easily reproducible with a fresh copy of the current code and running the linear wave test case while trying to output the variable uov. There is also a typo in the changes (line 688) so uov0 works but user_out_var0 does not. abbr_name is repeated twice but I think the intention was for the second condition to be ContainVariable(output_params.variable, full_name).

kahoooo:
## Prerequisite checklist

## Description

Closes #436 

## Testing and validation

Tested on my machine with problem `linear_wave` configured as followed:
`python configure.py --prob=linear_wave --cxx=clang++ --cflag='-arch arm64' -debug -hdf5`
Edited `athinput.linear_wave1d`, added hdf5 output and `,uov` to variable.
Code runs without producing error and the uov is output as intended.

### DIFF
--- a/src/outputs/outputs.cpp
+++ b/src/outputs/outputs.cpp
@@ -678,14 +678,17 @@ void OutputType::LoadOutputData(MeshBlock *pmb) {
     }
   } // endif (MAGNETIC_FIELDS_ENABLED)
 
+  bool output_all_uov = ContainVariable(output_params.variable, "uov")
+                        || ContainVariable(output_params.variable, "user_out_var");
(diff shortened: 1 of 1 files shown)


c-white (approved):
The logic looks good to me.

PR #444, opened 2022-05-18
Passive scalar dependent EOS

### LINKED ISSUE #359: Integrate passive scalars into the EOS
**Is your feature request related to a problem? Please describe.**
I want to extend the General EOS capabilities to include passive scalars. For example, one could have a composition dependant EOS and use the passive scalars to track composition.

My main purpose of this feature request is to gather input from others to help me implement this feature myself.

**Describe the solution you'd like**
The primary issue here is that the EOS is called within the Riemann Solver. Therefore the face reconstructed passive scalars need to be passed to the Riemann Solver, and then passed to the EOS calls within. Currently the reconstruction and the flux calculations for the hydro and scalar variables are completely separated with different function calls and different tasks. Ideally, these would be unified in some way.

The two main ways I see of doing this is either:

1. Extending the hydro athena array to include the scalars, so that the zeroth scalar index is `NHYDRO` (when MHD is disabled).
2. Keep the hydro and scalar arrays separated but combine the two separate flux/reconstruction functions/tasks.

**Describe alternatives you've considered**
A possible alternative would be to add scalar reconstruction to the hydro steps but to only use this computation for the Riemann problem and to disregard it and recompute it in the scalar task. This adds unnecessary computation but is probably least likely to break something.

**Additional context**
My knowledge of the task list is very poor and I would like need help modifying it.

msbc:
Add support for equations of state that depend on passive scalars. Closes #359 

## Prerequisite checklist

## Description

To enable support for equations of state that depend on passive scalars, the passive scalars must be passed to (nearly) all EOS calls. The main difficulty here is that EOS calls are made inside the Riemann solver. This requires that the face centered values of the scalar primitive values must also be passed to the Riemann solver. Which, in turn, requires that the scalars be interpolated to the face centers at the same time as the hydro variables. Thus `PasiveScalars::CalculateFluxes` was essentially merged into `Hydro::CalculateFluxes` (I have left `PasiveScalars::CalculateFluxes` in the code even-though it is unused). Otherwise I've done my best to keep the scalars as separated from hydro as I can. Although, I think we should combine the hydro and scalar `ConservedToPrimitive` and `PrimitiveToConserved` functions as they are always called together (outside of pgens and if `NSCALARS>0`). If your EOS needs to do an inversion (e.g. find Temperature) it would be useful to store this in the passive scalars to estimate the solution for the next EOS inversion.

Currently 4th-order is not properly supported for an EOS which depends on passive scalars. In `EquationOfState::ConservedToPrimitiveCellAverage` there is one call to ConservedToPrimitive that uses the 4th-order cell-centered values, but I am currently only passing it the cell-averaged scalars. The only way I see of fixing that is to merge `PasiveScalarPrimitiveCellAverage` into `ConservedToPrimitiveCellAverage` so that the 4th-order cell-centered scalars are accessible at this stage. However, I didn't want to make this change without approval from @felker.

You may want to categorize major vs. minor changes and list them:
1. Change nearly all EOS function and Riemann solver signatures to include passive scalars, enabling scalar dependent EOS.
2. Merged `PasiveScalars::CalculateFluxes` into `Hydro::CalculateFluxes`
3. The Riemann solvers are now responsible for computing the scalar fluxes
4. Removed `CalculateScalarFlux` from task list as it is now part of hydro fluxes (dependancies updated)

## Testing and validation

All current tests past using `--cxx=clang++-apple` on my MacBook Pro (with Apple Arm64 CPU).

...

## To-do

- [ ] Add EOS module that depends on passive scalars
- [ ] Add corresponding shock tube tests for regression testing
- [ ] Double check passive scalar fluxes are properly updated for SR/GR
- [ ] Make sure doxygen comments are updated
- [ ] Fix 4th-order
- [ ] Cleanup now unused scalar code (and `mass_flux_fc`)
- [ ] Maybe find a cleaner way of passing empty scalar arrays when `NSCALARS` is zero

### DIFF
--- a/src/bvals/bvals.cpp
+++ b/src/bvals/bvals.cpp
@@ -46,6 +46,9 @@
 #include <mpi.h>
 #endif
 
+// empty array to pass in place of scalars when NSCALARS=0
+static AthenaArray<Real> empty;
+
 //----------------------------------------------------------------------------------------
 //! \brief BoundaryValues constructor
 //!        (the first object constructed inside the MeshBlock() constructor)
@@ -488,8 +491,10 @@ void BoundaryValues::ApplyPhysicalBoundaries(const Real time, const Real dt,
                                               pmb->is-NGHOST, pmb->is-1,
                                               bjs, bje, bks, bke);
     }
-    pmb->peos->PrimitiveToConserved(ph->w, pf->bcc, ph->u, pco,
-                                    pmb->is-NGHOST, pmb->is-1, bjs, bje, bks, bke);
+    pmb->peos->PrimitiveToConserved(ph->w, pf->bcc, ph->u,
+                                    (NSCALARS) ? ps->r : empty,
+                                    (NSCALARS) ? ps->s : empty,
+                                    pco, pmb->is-NGHOST, pmb->is-1, bjs, bje, bks, bke);
     if (NSCALARS > 0) {
       pmb->peos->PassiveScalarPrimitiveToConserved(
           ps->r, ph->u, ps->s, pco, pmb->is-NGHOST, pmb->is-1, bjs, bje, bks, bke);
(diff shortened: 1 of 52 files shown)


msbc:
I recently picked this back up, but let's mark it as closed for now.

PR #445, opened 2022-05-27
Initialize pointer fv_new_ to nullptr

kahoooo:
Just stumble upon this bug that is only triggered for decaying turbulence (turb_flag == 1).
When `turb_flag == 1`, `fv_new_` would be pointing to random address and the destructor of `TurbulenceDriver` would attempt to access and free `fv_new_[nv]`.

### DIFF
--- a/src/fft/turbulence.cpp
+++ b/src/fft/turbulence.cpp
@@ -59,7 +59,8 @@ TurbulenceDriver::TurbulenceDriver(Mesh *pm, ParameterInput *pin) :
          {nmb, pm->my_blocks(0)->ncells3,
                pm->my_blocks(0)->ncells2, pm->my_blocks(0)->ncells1},
          {nmb, pm->my_blocks(0)->ncells3,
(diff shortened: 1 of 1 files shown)


src/fft/turbulence.cpp
```diff
@@ -93,6 +93,7 @@ TurbulenceDriver::TurbulenceDriver(Mesh *pm, ParameterInput *pin) :
   fv_ = new std::complex<Real>*[3];
   fv_sh_ = new std::complex<Real>*[3];
   fv_co_ = new std::complex<Real>*[3];
+  fv_new_ = nullptr;
```
changgoo, line 96: Can you instead add
```
fv_new_(null_ptr) {
```
at line 63?

changgoo (changes requested):
Thanks for catching this. It has been fixed in my private repo but forgot to fix it here.

src/fft/turbulence.cpp
kahoooo: Sure! Done.

changgoo (approved):

Issue #365, opened 2020-02-10
Passive scalar issue with AMR


morganemacleod:
Hi Folks, 

I have an issue that I first noticed in my own problem and have since been able to reproduce (though to a somewhat less extreme extent) in the blast with hydro. 

It stems from using adaptive refinement with passive scalars.  I think that when meshblocks are de-refined, their boundaries seem to introduce incorrect values of the scalar concentration. 

To reproduce I modified the blast problem by introducing a single passive scalar. I commented the changes to blast.cpp (attached) with "MM": [blast_scalar.cpp.txt](https://github.com/PrincetonUniversity/athena-public-version/files/4181814/blast_scalar.cpp.txt) (rename -> .cpp) 

I configure with: 
<code>
python configure.py --prob blast_scalar --coord spherical_polar --flux hllc -mpi -hdf5 --hdf5_path /opt/local --cxx g++ --nscalars=1
</code>

And I run with the following input file, which increases the number of AMR levels and changes the threshold for refinement vs the default. 
[athinput.blast_sph.txt](https://github.com/PrincetonUniversity/athena-public-version/files/4181840/athinput.blast_sph.txt)
(again, remove the .txt)

I read into a python notebook as follows:
```
import matplotlib.pyplot as plt
import numpy as np
import athena_read as ar

%matplotlib inline

d = ar.athdf("/Users/morganmacleod/DATA/athenaruns/blast/scalar/Blast.out1.00001.athdf")

plt.figure(figsize=(8,8))
th_ind = int(d['rho'].shape[1]/2)
rr,pp = np.meshgrid(d['x1v'],d['x3v'])
plt.pcolormesh(rr*np.cos(pp),rr*np.sin(pp),d['r0'][:,th_ind,:])
plt.colorbar(label='r')
plt.xlabel('x')
plt.ylabel('y')
plt.axis('equal')
```

To generate the following:
![download-6](https://user-images.githubusercontent.com/22401914/74167662-914cf800-4bf6-11ea-95c1-4722b9d25964.png)
What I see here, is that in the interior of the "bubble" where the mesh has de-refined, along the boundaries there are some artifacts introduced. 

When I ran this with my own problem (which unfortunately requires other customizations of athena) I was encountering values of r-> 1e20 ! But I haven't been able to reproduce that extreme behavior in the blast problem -- I thought that this, based on the unmodified public repo, was probably a more useful case. 

With a fixed mesh, for example, we see no artifacts in the scalar concentration:
[...]

PR #446, opened 2022-05-28
Fixed passive scalar + AMR bug

c-white:
I noticed something wrong with the passive scalars near refinement boundaries in an AMR calculation. After digging through the code, I realized the scalars were not perfectly paralleling the hydro quantities with regard to primitive vs. conserved coarse buffers. In particular, calls to `HydroBoundaryVariable::SwapHydroQuantity` were followed by the appropriate setting of `var_cc` in the scalars' `CellCenteredBoundaryVariable`, but the subsequent setting of `coarse_buf` done by `HydroBoundaryVariable::SelectCoarseBuffer` was omitted. This occurs 7 times in `mesh.cpp`, 4 times in `time_integrator.cpp`, and 2 times in `sts_task_list.cpp`.

This pull request adds in the missing logic in all 13 cases. I don't know if this is strictly necessary in all cases, but it doesn't seem like this can harm anything. I am particularly uncertain about STS, since I never use it.

This problem is hard to find in existing test problems. Partly, this may be because the effects of bad ghost cells may be hidden by flux corrections and time evolution. It may also not show up in problems where the conserved density is 1, since the primitive and conserved scalars are the same in that case. It may well affect SMR calculations, but I can't be certain I've seen that occur.

Below is an example problem before and after this update. The hydro variables always look good, but the passive scalars show clear grid artifacts in the unfixed version.

amr_scalar_comparison

- [x] Code follows style guide.
- [x] Documentation updated (N/A).
- [ ] Test coverage: I could not find a simple test that illustrated the problem (I only saw it with a 3D AMR multigrid calculation with at least 512 8^3 blocks).
- [ ] Optional: someone can comment on whether all the changes here are necessary.
- [ ] Optional: one could achieve the same effect by making a new `ScalarBoundaryVariable` derived from `CellCenteredBoundaryVariable` with analogous `SwapScalarQuantity` and `SelectCoarseBuffer` functions, more closely paralleling hydro.

### DIFF
--- a/src/mesh/mesh.cpp
+++ b/src/mesh/mesh.cpp
@@ -1431,8 +1431,13 @@ void Mesh::Initialize(int res_flag, ParameterInput *pin) {
         if (MAGNETIC_FIELDS_ENABLED)
           pmb->pfield->fbvar.SendBoundaryBuffers();
         // and (conserved variable) passive scalar masses:
-        if (NSCALARS > 0)
+        if (NSCALARS > 0) {
+          pmb->pscalars->sbvar.var_cc = &(pmb->pscalars->s);
+          if (pmb->pmy_mesh->multilevel) {
+            pmb->pscalars->sbvar.coarse_buf = &(pmb->pscalars->coarse_s_);
+          }
           pmb->pscalars->sbvar.SendBoundaryBuffers();
+        }
       }
 
       // wait to receive conserved variables
@@ -1470,6 +1475,9 @@ void Mesh::Initialize(int res_flag, ParameterInput *pin) {
           pmb->phydro->hbvar.SendBoundaryBuffers();
           if (NSCALARS > 0) {
             pmb->pscalars->sbvar.var_cc = &(pmb->pscalars->r);
+            if (pmb->pmy_mesh->multilevel) {
+              pmb->pscalars->sbvar.coarse_buf = &(pmb->pscalars->coarse_r_);
+            }
             pmb->pscalars->sbvar.SendBoundaryBuffers();
           }
         }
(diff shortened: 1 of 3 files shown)


pdmullen (commented):
It looks like @c-white may have just brought #365 to its grave.  

As far as STS goes, it should just invoke the exact same infrastructure as `time_integrator.cpp`, which looks like what you have done here.  Nice find and happy Memorial Day weekend 😸

tomidakn (approved):

felker:
This is an amazing find @c-white . The changes make sense to me (worst case, some of the swapping of coarse buffer variable references is redundant, but shouldnt be a problem). 

I am more surprised how they were completely missing from my original implementation. And why my design decisions did not end up with making a subfolder `src/bvals/cc/scalar` and associated `CellCenteredBoundaryVariable` derived class, like you suggest. I suppose I envisioned passive scalars as truly the generic concrete example of a `CellCenteredBoundaryVariable`, and by the time I got through the mess of writing https://github.com/PrincetonUniversity/athena/blob/master/src/bvals/bvals_refine.cpp, I didnt revisit that assumption. Also surprised how the failures were so small, although that could be attributed to only introducing problems at boundaries with **coarser** MeshBlocks (and the difference would be large only when density is far from 1.0?).

Certainly there are several unique characteristics of Hydro variables as boundary variables (mostly shearing box, orbital advection related), but this comment is definitely untrue and should be changed:
https://github.com/PrincetonUniversity/athena/blob/afcd297a9b2b8e91ae06c0637911d49f8eba081d/src/bvals/cc/hydro/bvals_hydro.hpp#L59-L60

In other words, much of https://github.com/PrincetonUniversity/athena/blob/master/src/bvals/cc/hydro/bvals_hydro.cpp and https://github.com/PrincetonUniversity/athena/blob/afcd297a9b2b8e91ae06c0637911d49f8eba081d/src/mesh/mesh_refinement.cpp#L1071 should be modified to have generic implementations that work for hydro variables and passive scalars, and any cell-centered variable that has some kind of notion of conserved and primitive variable duality that should be handled nearly the same way by AMR.

felker (approved):

PR #454, opened 2022-08-31
Apply output time cadence changes even if no input file is provided during restarts

### LINKED ISSUE #310: Output next time is ignored on startup

### LINKED ISSUE #451: Command-line output#/dt and output#/next_time are handled incorrectly
**Prerequisite checklist**

Place an X in between the brackets on each line as you complete these checks:
- [X] Did you check that the issue hasn't already been reported?
- [X] Did you check the documentation in the Wiki for an answer?
- [X] Are you running the latest version of Athena++?

**Summary of issue**

When `output#/dt` and `output#/next_time` are specified during a restart, fix-ups applied to output times (see #310) are omitted when these two parameters are on the command line rather than in an input file.

**Steps to reproduce**

We use two input files `athinput-time1` and `athinput-time2` containing, among others, the following lines:
```
<job>
problem_id = time
<time>
tlim = 10
<output1>
file_type = rst
dt = 1
```
```
<job>
problem_id = time
<time>
tlim = 10
<output1>
file_type = rst
dt = 0.5
next_time = 1.8
```

**Expected and actual outcomes**

If we run Athena++ with
```bash
mkdir time{1,2}
bin/athena -d time1 -i athinput-time1
bin/athena -d time2 -i athinput-time2 -r time1/time.00005.rst
```
and we read the binary timestamps off the restart files, we get
```
time1/time.00000.rst 0.0
time1/time.00001.rst 1.0000092775033713
time1/time.00002.rst 2.0003549602099246
time1/time.00003.rst 3.0008885306036044
time1/time.00004.rst 4.000559354421319
time1/time.00005.rst 5.000100787631505
time1/time.00006.rst 6.000578975742493
time1/time.00007.rst 7.001832959788023
time1/time.00008.rst 8.000395785266205
time1/time.00009.rst 9.001636920101932
time1/time.final.rst 10.0
```
```
time2/time.00006.rst 5.300735126408735
time2/time.00007.rst 5.8014517407979715
time2/time.00008.rst 6.300726294886482
time2/time.00009.rst 6.801709105815053
time2/time.00010.rst 7.301515963883339
time2/time.00011.rst 7.800602370859581
time2/time.00012.rst 8.30158045147817
time2/time.00013.rst 8.801768346195535
time2/time.00014.rst 9.30166771908827
time2/time.00015.rst 9.800298307581457
time2/time.final.rst 10.0
```
That is to say, the next output after the restart happens at time `next_time + n * dt` that is later than the restart time for some non-negative integer `n`. I would argue the output should happen as soon as possible if `next_time` is earlier than the restart time, because that would give us an easy way to make a debug dump right at the point of restart, which would be useful if the restart file has been modified externally, or if the restart process is customized to do more than just reading in the data, but that parameter interpretation is up for debate.

A more serious problem arises if we then run
```bash
rm time2/*
bin/athena -d time2 -r time1/time.00005.rst output1/dt=0.5 output1/next_time=1.8
```
The timestamps are now
```
time2/time.00006.rst 5.002184280517581
time2/time.00007.rst 5.004198237047655
time2/time.00008.rst 5.006251402013058
time2/time.00009.rst 5.008233401549724
time2/time.00010.rst 5.010171844156113
time2/time.00011.rst 5.012137908737955
time2/time.00012.rst 5.0140426303780625
time2/time.00013.rst 5.300735126408735
time2/time.00014.rst 5.8014517407979715
time2/time.00015.rst 6.300726294886482
time2/time.00016.rst 6.801709105815053
time2/time.00017.rst 7.301515963883339
time2/time.00018.rst 7.800602370859581
time2/time.00019.rst 8.30158045147817
time2/time.00020.rst 8.801768346195535
time2/time.00021.rst 9.30166771908827
time2/time.00022.rst 9.800298307581457
time2/time.final.rst 10.0
```
Output happens and `next_time += dt` every cycle until `next_time` exceeds the simulation time. The utility of this catch-up is almost zilch, unless the user explicitly wants `next_time == file_number * dt`. By the way, `output#/file_number` actually refers to the next output, in the same way `next_time` does; that is not documented in the wiki.

However, if we run:
```bash
rm time2/*
bin/athena -d time2 -i athinput-time2 -r time1/time.00005.rst output1/dt=0.7 output1/next_time=1.9
```
The timestamps are normal:
```
time2/time.00006.rst 5.40105071040028
time2/time.00007.rst 6.100257532253666
time2/time.00008.rst 6.801709105815053
time2/time.00009.rst 7.500868906510636
time2/time.00010.rst 8.201493737794491
time2/time.00011.rst 8.901415861600112
time2/time.00012.rst 9.601331998722616
time2/time.final.rst 10.0
```
Note that the parameters on the command line are changed to demonstrate that they do override the ones in the input file. The order of command-line arguments is immaterial.

The examples here use restart files for simplicity, but the same bug should apply to other forms of output as well.

**Additional comments and/or proposed solution**

There have been multiple issues addressing time-step issues: #36, #62, #116, #310. #62 mentions the function [ParameterInput::ForwardNextTime()](https://github.com/PrincetonUniversity/athena/blob/4c26f709f71dbabf1d8fb76f2ee1195b5c8ec006/src/parameter_input.cpp#L789); it seems that this function and its friend [ParameterInput::RollbackNextTime()](https://github.com/PrincetonUniversity/athena/blob/ae02dd19c3510cf4c170d2d22ef80d3979c4f2c5/src/parameter_input.cpp#L753) are called only when an input file is supplied:

https://github.com/PrincetonUniversity/athena/blob/5871884c40f3d93d11bb8bd447fd03506ba24711/src/main.cpp#L229-L231
https://github.com/PrincetonUniversity/athena/blob/5871884c40f3d93d11bb8bd447fd03506ba24711/src/main.cpp#L297-L301

**Version info**

- Athena++ version: v21.0-dev-121-gae02dd19
- Compiler and version: g++ (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
- Operating system: Ubuntu 20.04.4 LTS
- Hardware and cluster name (if applicable):
- External library versions (if applicable): openmpi-bin:amd64 (4.0.3-0ubuntu1), libhdf5-openmpi-dev:amd64 (1.10.4+repack-11ubuntu1)

felker:
Makes command line changes equivalent to input file changes. Closes #451. Can you check @EdwinChan ? I just removed the `if (iarg_flag == 1)` check for the `Forward/RollbackNextTime()` calls in `main.cpp`, which I am pretty sure is safe in all cases but not 100% positive. 

Also closes #310, by skipping the initial condition output if `next_time > start_time`. This was one of two suggested changes; see discussion in #311. The alternative is to stop overriding `next_time` when the IC is outputted, in other words "keep `next_time` after the initial output."
- [ ] Document this change

---

#451 had many other comments and suggestions, which should be opened as separate issues if we want to pursue any of them:
- [ ] Expose `outputN/file_number` as a user-modifiable options, like `outputN/next_time`:
```
./outputs/outputs.cpp:137:      op.next_time = pin->GetOrAddReal(op.block_name,"next_time", pm->time);
```
- [ ] Give the user the ability to overwrite `mesh->time` (I suppose this only matters for restarts, since you can change `time/start_time` for fresh simulations)
- [ ] Policy change about `next_time < pm->time` behavior when `next_time` is changed via input file and/or command line during restart. I regard this as an invalid/ill-defined choice that the user has no reason to assume will result in a particular behavior , so I think the current implementation is fine. Currently we use `ForwardNextTime()` to set "the next output after the restart happens at `next_time_new = next_time + n * dt` so that `next_time_new >= pm->time` for the smallest such non-negative integer `n`"

The suggestion is that "the output should happen as soon as possible ..., because that would give us an easy way to make a debug dump right at the point of restart", i.e. `next_time_new = pm->time` like an initial condition output for the restarted simulation, which does seem useful. 

Regardless of the choice, we should explicitly document it in the Wiki if it isnt already documented.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -226,9 +226,10 @@ int main(int argc, char *argv[]) {
     if (res_flag == 1) {
       restartfile.Open(restart_filename, IOWrapper::FileMode::read);
       pinput->LoadFromFile(restartfile);
-      // If both -r and -i are specified, make sure next_time gets corrected.
+      // make sure next_time gets corrected in case -i input file or cmdline args change
+      // the output next_time, dt, etc.
       // This needs to be corrected on the restart file because we need the old dt.
-      if (iarg_flag == 1) pinput->RollbackNextTime();
+      pinput->RollbackNextTime();
       // leave the restart file open for later use
     }
     if (iarg_flag == 1) {
@@ -295,8 +296,9 @@ int main(int argc, char *argv[]) {
 #endif // ENABLE_EXCEPTIONS
 
   // With current mesh time possibly read from restart file, correct next_time for outputs
-  if (iarg_flag == 1 && res_flag == 1) {
-    // if both -r and -i are specified, ensure that next_time  >= mesh_time - dt
+  if (res_flag == 1) {
+    // ensure that next_time  >= mesh_time - dt, in case input file or command line
+    // overrides it
     pinput->ForwardNextTime(pmesh->time);
   }
 
--- a/src/outputs/outputs.cpp
+++ b/src/outputs/outputs.cpp
@@ -776,10 +776,13 @@ void OutputType::ClearOutputData() {
 //! \brief scans through singly linked list of OutputTypes and makes any outputs needed.
 
 void Outputs::MakeOutputs(Mesh *pm, ParameterInput *pin, bool wtflag) {
+  // wtflag = only true for making final outputs due to signal or wall-time/cycle/time
+  // limit. Used by restart file output to change suffix to .final
   bool first=true;
   OutputType* ptype = pfirst_type_;
   while (ptype != nullptr) {
-    if ((pm->time == pm->start_time)
+    if (((pm->time == pm->start_time) // output initial conditions, unless next_time set
+         && (ptype->output_params.next_time <= pm->start_time ))
       || (ptype->output_params.dt > 0.0 && pm->time >= ptype->output_params.next_time)
       || (ptype->output_params.dcycle > 0 && pm->ncycle%ptype->output_params.dcycle == 0)
       || (pm->time >= pm->tlim)
--- a/src/outputs/restart.cpp
+++ b/src/outputs/restart.cpp
@@ -31,7 +31,8 @@
 
 
 //----------------------------------------------------------------------------------------
-//! \fn void RestartOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, bool flag)
+//! \fn void RestartOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin,
+//                                          bool force_write)
 //! \brief Cycles over all MeshBlocks and writes data to a single restart file.
 
 void RestartOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, bool force_write) {
@@ -46,9 +47,10 @@ void RestartOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, bool force_wr
 
   fname.assign(output_params.file_basename);
   fname.append(".");
-  // add file number to name, unless write is forced by terminate signal, in which case
-  // replace number in the name by the string "final".  This keeps the restart file
-  // numbers consistent with output.dt when a job is restarted many times.
+  // add file number to name, unless write is forced by signal or main integration loop,
+  // (wall-time / cycle / time limit) in which case replace number in the name by the
+  // string "final".  This keeps the restart file numbers consistent with output.dt when a
+  // job is restarted many times.
   if (!force_write)
     fname.append(number);
   else


EdwinChan:
Thanks, @felker, for the patch. The four tests in the original post of #451 have been redone with commit 08a002a0f057a390d320e3205d1615889f187d46; only the third test produces different results, and the new results are identical to the second test. That is to say, the specific problem I raised seems to be fixed. I'll let the maintainers contemplate the wider ramifications, if any.

I second the idea of documenting in the wiki the behavior when `next_time < pm->time` (as C++ programmers I'm sure we all hate undefined behavior…). Better yet, Athena++ can simply terminate on invalid input. As a bonus, we can teach the readers how to read the time in the restart file (see, e.g., https://github.com/PrincetonUniversity/athena/issues/451#issuecomment-1232344719) and set `output#/next_time` to that.

felker:
Thanks for checking @EdwinChani. Can you draft an edit to the Wiki, and I will apply it?

felker:
Any suggested edit to the Wiki, @EdwinChan ?

EdwinChan:
I felt like I'm being gaslighted by GitHub… I was away during the long weekend and I can clearly remember replying to this thread after I came back to the office last Wednesday, but that post just disappeared. It's giving me serious trust issues…

Anyway, here's my suggested edit to the [wiki](https://github.com/PrincetonUniversity/athena/wiki/The-Input-File):

- `next_time` (optional) : earliest time the simulation must reach before the next output; the behavior is undefined if `next_time` is earlier than `time/start_time` for an initial run (see `<time>` block below; default: `time/start_time`), or the time stored in the restart file for a restart run
- `file_number` (optional) : number by which the next output is labeled (default: 0; should satisfy 0 ≤ `file_number` ≤ 99999)

Some notes:

- "reach" means output is performed also when `pm->time == next_time`; "exceed" would exclude the equality case
- the behavior is still defined for the equality case, in which case the output should happen as soon as possible
- the way the main loop is currently written makes adhering to the specification laid out above difficult:
  - for initial runs, there's additional logic before the main loop that dumps the IC, but no time check is performed
  - for restart runs, the first output opportunity is after one iteration of the main loop, so the time is already different
- please check that my interpretation of the code gives the correct default value for `output#/next_time`
- the hint about the range of `output#/file_number` is to avoid formatting issues

felker:
https://github.com/PrincetonUniversity/athena/wiki/The-Input-File

how does this look?

EdwinChan:
The only quibble I have is that the phrase "to write the initial condition" implies that any other value of `next_time` would prevent writing of the initial condition, but in fact the initial condition is written unconditionally (pun?):

https://github.com/PrincetonUniversity/athena/blob/15bd136cd5237980965ce774d3ce4c5f1531980c/src/main.cpp#L399

I'd suggest removing that phrase.

felker:
Not anymore: https://github.com/PrincetonUniversity/athena/commit/13ebdb9cf3ce23e64a7cae11f5540de7c52d3c33

https://github.com/PrincetonUniversity/athena/blob/15bd136cd5237980965ce774d3ce4c5f1531980c/src/outputs/outputs.cpp#L778-L790

felker:
maybe good to add a comment in `main.cp` since the name of the function call is misleading

EdwinChan:
For what it's worth, setting `next_time` to the time in the restart file means the next output will happen at `next_time` + `dt`, necessitating "exceed" in the description. Immediate output after restart is possible if `next_time` + `dt` is slightly later than the restart time.

felker:
Are you suggesting I modify this?

>   * `next_time` (optional)  : earliest time the simulation must reach (or exceed) before the next output is written; the behavior is undefined if `next_time` is earlier than either 1) `time/start_time` for an initial, non-restart run 2) or the time stored in the restart file for a restart run (default = `time/start_time` to write the initial condition for a non-restart run)

EdwinChan:
Not really; the way you wrote it is loose enough to cover the situation I described. But someone years later might have Googled this page when trying to figure out why there is no output even though `start_time` is after the restart time, and I thought it might be worthwhile to mention it here.

PR #456, opened 2022-09-26
Fix Shearing Box + STS Hang

pdmullen:
# Summary
STS only sends/receives a subset of variables and fluxes depending on which diffusive physics are enabled.  For example, for an isothermal calculation invoking STS for Ohmic diffusion, only field variables/fluxes are communicated inside the `SuperTimeStepTaskList`.

When a shearing box overhaul in PR #342 made STS compatible with shearing box, this communication strategy was not reflected in the shearing box specific codebase.  @zhuzh1983 reported a hang for a calculation using an (1) isothermal EOS, (2) ambipolar diffusion, (3) STS, and (4) MPI.   I find that the changes in this PR remove the deadlock.  

Specifically, after PR #342, the function `StartReceivingShear` was not modified to reflect the aforementioned communication design: https://github.com/PrincetonUniversity/athena/blob/master/src/bvals/bvals.cpp#L361-L406.  Instead, all communicators were instantiated for hydro and field quantities (see the `for` loop over `bvars_main_int` on line 391).  I have changed how the `StartReceivingShear` function is handled so that receives are only instantiated for boundary variables contained in `bvars_sts` when `StartReceivingSubset` is called from the `SuperTimeStepTaskList`.  

# To-Do
I am rather unfamiliar with the shearing box bvals framework, therefore, I would greatly appreciate another set of eyes on this PR.  For example, were any implicit assumptions made elsewhere in the shearing box framework that assumed that all variables would be communicated @tomo-ono?  In other words, is this PR going to break shearing box calculations elsewhere? My testing is limited to finding a solution that fixes @zhuzh1983's deadlock.  Are there other tests available that someone can run to make sure everything works as anticipated?

Also worth bringing up: for STS, we introduced both  `StartReceivingSubset` and `ClearBoundarySubset` functions.  With shearing box, a special clause was added to the `StartReceivingSubset` function (and was modified in this PR).  However, there is no related shearing box specific clause in `ClearBoundarySubset`.  I assume this is because all communications instantiated in `StartReceivingShear` are cleared in `ClearBoundary` (rather than having a dedicated `ClearBoundaryShear` function).  This is confusing.... Just want to make sure that everything is okay here too, @tomo-ono.  

## Prerequisite checklist

### DIFF
--- a/src/bvals/bvals.cpp
+++ b/src/bvals/bvals.cpp
@@ -353,54 +353,26 @@ void BoundaryValues::StartReceivingSubset(BoundaryCommSubset phase,
   // KGF: begin shearing-box exclusive section of original StartReceivingForInit()
   // find send_block_id and recv_block_id;
   if (shearing_box != 0) {
-    StartReceivingShear(phase);
-  }
-  return;
-}
-
-//----------------------------------------------------------------------------------------
-//! \fn void BoundaryValues::StartReceivingShear(BoundaryCommSubset phase)
-//! \brief initiate MPI_Irecv() for shearing box
-//!
-//! \note
-//! - cannot simply combine StartReceivingShear() at end of StartReceiving()
-//!   (which is done for ClearBoundary), because the "shared"/non-virtual fn
-//!   BoundaryValues::FindShearBlock() must be called in between 2x fns
-//! - shearing box is currently incompatible with both GR and AMR
-void BoundaryValues::StartReceivingShear(BoundaryCommSubset phase) {
-  switch (phase) {
-    case BoundaryCommSubset::mesh_init:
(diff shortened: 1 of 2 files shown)


pdmullen:
@tomo-ono  @felker @tomidakn @zhuzh1983, would you mind reviewing this?

tomo-ono:
Processes of shear periodic boundary for hydro, field, and passive scalars should be independent each other. Therefore, this change should not break shearing box processes. As @pdmullen points out, defining `ClearBoundaryShear` would leads better designed. I will have another PR for it after this PR is merged.

felker:
LGTM, any objections to merging?

buildbot-princeton:

Issue #430, opened 2022-03-08
"hanging" simulations 


nchoksi4456:
**Summary of issue**

My simulation "hangs" at a seemingly random point -- it suddenly starts to take inordinately long to run (see the plot below). The issue only crops up for certain choices of resolution. I'm running pure hydro to study planet-disk interactions. Static mesh refinement is included near the planet. 

Any ideas people have for debugging this issue would be appreciated (and sorry if this ends up having a simple solution).

**Steps to reproduce**

My problem generator can be found [here](http://ugastro.berkeley.edu/~nchoksi/zhu_disk_original.cpp) and my input file [here](http://ugastro.berkeley.edu/~nchoksi/zhu_input_h3.500_qth_0.1).

**Actual outcome**
The top panel of the plot below shows the simulation timestep for two different runs. The pink curve is a lower resolution model that runs to completion without a hitch. The black curve is a higher resolution (in the static mesh refinement region) model.  In both runs, the simulation timestep gradually decreases as the simulation progresses.

The bottom panel of the plot shows the actual walltime between each 0.01 of simulation time. For a while this walltime increases commensurately with the decreasing timestep. But then at t ~ 35, the simulation "hangs" -- the walltime between successive 0.01 intervals of simulation time increases drastically, despite the fact that the simulation timestep is behaving reasonably. So the code must be spending a lot of time on something else. Not sure what.

<img width="301" alt="hanging" src="https://user-images.githubusercontent.com/25126947/157285278-5629deb5-204b-4d11-b259-d5999d97e025.png">

The problematic/hanging simulations finish running without a hitch if I restart them from a checkpoint. 

Also, I've checked that this is not a one-off issue. I ran the setup a few times using different nodes (though all on the same cluster). 

Thanks!

**Version info**

- Athena++ version: v21.0
- Compiler and version:  gcc (GCC) 6.3.0
- Operating system:  Scientific Linux 7
- Hardware and cluster name (if applicable):
- External library versions (if applicable):

felker:
What is your full configure + compilation command? Just to be clear, you are doing this on a single CPU core?

When you Ctrl-C SIGINT the code, does it tell you what function it was in when the signal was caught? Have you tried using gprof or even just print statements to see where it gets stuck?

nchoksi4456:
> What is your full configure + compilation command?

`python configure.py --prob zhu_disk_original --coord spherical_polar --eos isothermal -mpi -hdf5`


[...]

PR #458, opened 2022-10-11
fix a bug in disk.cpp

tomo-ono:
As pointed out in #430, disk.cpp has a bug. 
This PR just modifies the bug, `x3v(i)` -> `x3v(k)`.


tomidakn (approved):

buildbot-princeton:

PR #461, opened 2022-10-11
Make turbulent velocity field resolution independent

sanghyukmoon:
## Prerequisite checklist

## Description

When `rseed >= 0`, the turbulent power spectrum is generated with a global random number sequence, such that the perturbed velocity field is independent of the Meshblock configuration. However, current implementation does not give identical perturbation when the total number of cells (i.e., Mesh size) is changed. For example, the below figure plots the velocity perturbations in z=0 plane initialized with `rseed = 0`, using 32^3 (upper panels) and 64^3 (lower panels) Mesh (using `src/pgen/turb.cpp`), showing that the turbulent velocity field is different for different Mesh size.
original

This can be fixed by drawing random numbers only when the wavenumber is in the range specified by `nlow` and `nhigh`, otherwise the global random number sequence would depend on the maximum possible wavenumber set by the number of cells in each dimension. After applying the current PR, the velocity perturbation using `rseed >= 0` will not depend on the number of cells in the Mesh, as seen in the figure below.
fixed

This feature will be useful to make identical initial conditions for resolution study.

### DIFF
--- a/src/fft/turbulence.cpp
+++ b/src/fft/turbulence.cpp
@@ -258,27 +258,36 @@ void TurbulenceDriver::PowerSpectrum(std::complex<Real> *amp) {
     for (int gk=0; gk<kNx3; gk++) {
       for (int gj=0; gj<kNx2; gj++) {
         for (int gi=0; gi<kNx1; gi++) {
+          std::int64_t nx = GetKcomp(gi, 0, kNx1);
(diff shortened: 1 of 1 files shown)


changgoo (approved):
Looks good to me, and it is a useful feature.

tomidakn (approved):

PR #489, opened 2023-03-23
Chemistry update units

changgoo:
## Prerequisite checklist

## Description

Redefine variable names used in the Units class to conform the style guide (lower case only) and be more explicit.

## To-do

- [ ] @munan I assume `chemistry/utils/chemistry_utils.hpp` is obsolete, which has its own unit definition.
- [ ] Currently, the explicit unit definition is only coupled with chemistry so that the `Units` class is constructed in the `ChemNetwork` class. It might make more sense to setup the `Units` class in the Mesh constructor in the future.


munan:
- add CUBE definition in defs.hpp.in
- add units/*.cpp in Makefile.in
- added exception header in unit.cpp
- deleted obsolete unit definitions and functions in chemistry_utils.cpp
- Moved unit class construction to meshblock

Issue #545, opened 2023-09-11
Time Step Incorrect for Some Metrics in GR


smressle:
**Summary of issue**

The time steps in src/hydro/new_blockdt.cpp used for GR could potentially be incorrect when the coordinate speed of light in a given direction differs much from 1.  Currently, the minimum time step is set by (e.g., for the x1 direction)

`pmb->pcoord->CenterWidth1(k, j, is, ie, dt1);`

where CenterWidth returns sqrt(g(I11) dx1

However, this is clearly incorrect when you take a coordinate transformation t' = t/10  x' = x, the metric remains unchanged except g(I00) gets multiplied by 100.  So the timestep used in Athena++ would remain unchanged even though the time coordinate is now a factor of 10 smaller.   So I think the correct thing to do is to calculate the coordinate speed of light for a given direction, e.g.:

coordinate_c = ( -g(I01) + std::sqrt( SQR(g(I01)) - g(I01)*g(I11) ) )/ ( g(I11) ) ; 

dt1(i) /= coordinate_c

**Steps to reproduce**
Use gr_user coordinates and set g(I00) = -10, g(I11) = 1, g(I22) = 1, g(I33) = 1

**Expected outcome**
The code will be numerically unstable for some problems

e.g., you can get things that look like this:

<img width="579" alt="Screen Shot 2023-09-08 at 11 07 33 PM" src="https://github.com/PrincetonUniversity/athena/assets/8203846/ae097d1c-dc85-4a03-9ba5-1081bfb91536">

c-white:
Finally getting around to addressing this...

First, I agree that the existing scheme has the problems identified.

Next, see where the motivation for the quadratic equation suggested, but I worry it doesn't work in general. Consider the theta-direction in SKS coordinates. The argument of the square root will be negative inside the ergosphere, and then what does one do? The equation was derived assuming $k^\mu$ has only a time component and a single space component. Even light cannot propagate purely in the $\theta$-direction inside the ergosphere, but it can propagate in the $\theta$-direction if it also moves prograde.

[...]

Issue #505, opened 2023-05-16
Missing documentation and instructions for creating user-defined Coordinates class; Get/Compute naming convention is confusing


felker:
Brought up by Lizhong Zhang at the 2nd Athena++ Workshop. This page on the Wiki: https://github.com/PrincetonUniversity/athena/wiki/Coordinate-Systems-and-Meshes should be expanded with another section (although it is already quite long) about the code modifications that are necessary for adding a new (relativistic and/or nonrelativistic?) Coordinates derived class to Athena++. Should probably include:
- Adding a new `.cpp` file to `coordinates/` directory
- Modifying `coordinates.hpp`
- Need to modify:https://github.com/PrincetonUniversity/athena/blob/99abbff15962419d773ee3f830c03cc8f6a8d914/configure.py#L335 if the new coordinate system is not GR/SR
- Virtual functions that should be overridden; class arrays that should be filled. 
- What needs to be precomputed and stored, and what functions are just for convenience. 
- Why we decided not to precompute and store a bunch of quantities like length, area, volume in Coordinates class arrays?

Anything else?

------------------------------------------
Also, related:

Our Coordinates abstract base class has 3x types of methods that are "Get"-prefixed counterparts to existing methods:
```
Real GetEdgeXLength(const int k, const int j, const int i);
...
Real GetFaceXArea(const int k, const int j, const int i);
...
Real GetCellVolume(const int k, const int j, const int i);
```
The non-"Get" functions fill a provided `AthenaArray` at a given `k,j` along an `x1` slice. These functions imply that they are "getters" for accessing private members/arrays of the Coordinates class, but they actually compute the quantities like their counterparts, just at a single cell.

A user/developer may have no idea why there are seemingly 2x versions of the same type of function (including me, I completely forgot):
- The `GetEdgeXLength()` functions seem to only be used in mesh refinement flux corrections?
- Same for the  `GetFaceXLength()` functions, but they are also used in the GR Torus problem generator. 
- `GetCellVolume()` is only used in a few problem generators

- [ ]  Do we rename the Get functions?

c-white:
Just to clarify the relativistic/nonrelativistic issue for anyone reading this thread: This task is about writing new Coordinates classes deep inside the code, modifying `src/coordinates/` in the process. Such coordinates might be used for any physics.

[...]

Issue #498, opened 2023-05-10
`athena_read.py` produces corrugated spherical-polar `x2v` coordinates due to float precision at high resolution


astroboylrx:
**Prerequisite checklist**

Place an X in between the brackets on each line as you complete these checks:
- [X] Did you check that the issue hasn't already been reported?
- [X] Did you check the documentation in the Wiki for an answer?
- [X] Are you running the latest version of Athena++?

**Summary of issue**

When reading data from a spherical-polar simulation with SMR, `athdf` in `athena_read.py` produces corrugated `x2v` coordinates. I found this when `scipy.interpolate.RegularGridInterpolator` reported that `The points in dimension 1 must be strictly ascending or descending`.

Specifically, `data[xf] = np.linspace(xmin, xmax, nx + 1, dtype=dtype)` at [here](https://github.com/PrincetonUniversity/athena/blob/fe6561b10e0abae60c75b60bdc0039119941d1f0/vis/python/athena_read.py#L648) and `data[xv] = np.empty(nx, dtype=dtype)` ([here](https://github.com/PrincetonUniversity/athena/blob/fe6561b10e0abae60c75b60bdc0039119941d1f0/vis/python/athena_read.py#L656)) make the face/center coordinates inherit the `dtype` of the first variable ([here](https://github.com/PrincetonUniversity/athena/blob/fe6561b10e0abae60c75b60bdc0039119941d1f0/vis/python/athena_read.py#L407)), which makes the `center_func` (here) unable to produce correct center coordinates as the algorithm is sensitive to floating precision.

The picture below shows that the `x2v` (288 coordinates between (xmin, xmax) = (1.5271630734205246, 1.614429548382759)) is corrugated, where the differences between consecutive coordinates sometimes become negative:
![Figure 8](https://github.com/PrincetonUniversity/athena/assets/3999966/292767ac-b486-4448-9449-b6ecf6511341)

Since the keyword `dtype` also controls the precision of simulation data, it is not desired to specify `dtype=np.float64` when calling `athdf`. 

---
**A quick easy fix** would be removing all the `dtype=dtype` in [this for loop](https://github.com/PrincetonUniversity/athena/blob/fe6561b10e0abae60c75b60bdc0039119941d1f0/vis/python/athena_read.py#L615). So the coordinates will fall back to the default `np.float64` precision. 

**Version info**

[...]

Issue #475, opened 2023-01-21
AMR in GR Does Not Work (Undefined Behavior)


smressle:
**Prerequisite checklist**

Place an X in between the brackets on each line as you complete these checks:
- [X] Did you check that the issue hasn't already been reported?
- [X] Did you check the documentation in the Wiki for an answer?
- [X] Are you running the latest version of Athena++?

**Summary of issue**

Running the code with GR and AMR without magnetic fields produces artificial clumping behavior in density.  Running it with MHD but B=0 everywhere initially fixes that clumping behavior in density but the field becomes nonzero in the same pattern. This seems to happen with MPI and AMR turned on in 3D.

e.g., this is bsq after some time (BH is in the center). AMR levels are concentrated around x=0, z=20rg, where x is horizontal and z is vertical.
<img width="578" alt="Screen Shot 2023-01-10 at 6 57 45 PM" src="https://user-images.githubusercontent.com/8203846/213881246-9b79ca0f-9346-484d-9bae-fe6f06f53335.png">

**Steps to reproduce**
Run a simple bondi-type flow (uniform initial conditions with B=0) with a refinement condition that results in refinement of at least two levels (not sure if this is required).  By the first time step nonzero fields arise.

You may want to include small snippets of Athena++ source code to describe the problem in more detail:

```
# Terminal output may be pasted here
```
**Actual outcome**

You may attach a plot here, if it is appropriate.

**Additional comments and/or proposed solution**
I have tracked down the issue to these two functions (and the 4D versions as well)

```c++
namespace BufferUtility {
//----------------------------------------------------------------------------------------
//! \fn template <typename T> void PackData(const AthenaArray<T> &src, T *buf,
//!     int sn, int en, int si, int ei, int sj, int ej, int sk, int ek, int &offset)
//! \brief pack a 4D AthenaArray into a one-dimensional buffer

template <typename T> void PackData(const AthenaArray<T> &src, T *buf,
         int sn, int en, int si, int ei, int sj, int ej, int sk, int ek, int &offset) {
  for (int n=sn; n<=en; ++n) {
    for (int k=sk; k<=ek; k++) {
      for (int j=sj; j<=ej; j++) {
#pragma omp simd
        for (int i=si; i<=ei; i++)
          buf[offset++] = src(n,k,j,i);
      }
    }
  }
}

//----------------------------------------------------------------------------------------
//! \fn template <typename T> void PackData(const AthenaArray<T> &src, T *buf,
[...]

PR #549, opened 2023-12-06
Miscellaneous small fixes

c-white:
Various issues have now been addressed:

- [x] Removing erroneous error reported by code when attempted to use SMR with 2D polar_wedge boundaries.
- [x] Improving approximate timestep estimation for GR (#545)
- [x] Improve documentation of GR user-specified coordinate (#505)
- [x] Address Python reader using single-precision to emulate code's double-precision coordinates calculation (#498)
- [x] Looking in to AMR + GR (#475)
- [x] Figure out what's going on with `uniform.py` (#383)


tomidakn (approved):

felker (approved):
Fixes all look good to me. I might suggest adding a comment about how the `xf` and `xv` arrays are always `float64` and warn about potential type promotions for any mixed precision calculations involving these arrays (as you point out in https://github.com/PrincetonUniversity/athena/issues/498#issuecomment-1874501914)--- but it isnt 100% necessary.

PR #566, opened 2024-02-29
Fix x1v for Curvilinear+PPM

forrestglines:
Here's a small fix to what appears to be an unintentional bug when using curvilinear coordinates with PPM reconstruction.

In `src/reconstruct/reconstruction.cpp`, a stray reference leads to changing the cell-volume-averaged radius (`x1v`) to the geometric cell-midpoint radius ( halfway between adjacent `x1f`s) when reconstruction variables are pre-computed. 

I cleaned up these lines a bit since I was digging around there.

### DIFF
--- a/src/reconstruct/reconstruction.cpp
+++ b/src/reconstruct/reconstruction.cpp
@@ -331,9 +331,9 @@ Reconstruction::Reconstruction(MeshBlock *pmb, ParameterInput *pin) :
               // Mignone equation 23: simplification of entries in \beta matrix when
               // Jacobian is a simple power of the coordinate, e.g. radial coordinate:
               Real coeff = (m_coord + 1.0)/(col + m_coord + 1.0);
-              int pow_num = col + m_coord + 1;
-              int pow_denom = m_coord + 1;
-              int s = row - kNrows/2;  // rescale index from -2 to +1
+              const int pow_num = col + m_coord + 1;
+              const int pow_denom = m_coord + 1;
+              const int s = row - kNrows/2;  // rescale index from -2 to +1
               // initializing transpose:
               beta[col][row] =
                   coeff*(
@@ -382,7 +382,7 @@ Reconstruction::Reconstruction(MeshBlock *pmb, ParameterInput *pin) :
           }
         } else { // if (uniform[X1DIR]) {
           // Mignone section 2.2: conservative reconstruction from volume averages
-          Real io = std::abs(i - pmy_block_->is);  // il=is-1 ---> io = 2
+          const Real io = std::abs(i - pmy_block_->is);  // il=is-1 ---> io = 2
           // Notes:
           // - io (i offset) must be floating-point, not integer type. io^4 and io^8 terms
           // in below lines quickly cause overflows of 32-bit and 64-bit integer limtis in
@@ -390,7 +390,7 @@ Reconstruction::Reconstruction(MeshBlock *pmb, ParameterInput *pin) :
           // - io=1 should correspond to "is" (first real cell face)
           // - take absolute value to handle lower x1 ghost zone cell faces properly
           if (std::strcmp(COORDINATE_SYSTEM, "cylindrical") == 0) {
-            Real delta = 120.0*SQR(SQR(io)) - 360.0*SQR(io) + 96.0;
+            const Real delta = 120.0*SQR(SQR(io)) - 360.0*SQR(io) + 96.0;
             // Mignone equation B.9:
             // w_im1
             c1i(i) = -(2.0*io - 3.0)*(5.0*SQR(io)*io + 8.0*SQR(io) - 3.0*io - 4.0)/delta;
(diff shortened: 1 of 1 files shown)


buildbot-princeton:

felker:
Thanks @forrestglines. I am willing to merge, but I noticed that there are a handful of other (seemingly improper) references to coordinate terms in that file. I am guessing I added these in the rush before my PhD defense in summer 2019, so totally plausible that they are all mistakes

... still, I would like to think it through and remove all the references in this PR if it is appropriate

felker:
Looks like the references first appeared in 9f7e315. Maybe just inherited from that. There is no legitimate reason to use references here, right?

forrestglines:
> Looks like the references first appeared in [9f7e315](https://github.com/PrincetonUniversity/athena/commit/9f7e315436046b7b1ca9d2499f3fec032ec077d6). Maybe just inherited from that. There is no legitimate reason to use references here, right?

Perhaps for performance they should be (`const`) references, although with modern compilers I doubt there would be a performance difference.

c-white:
I would also expect, though I don't know, that (preferably const) references to have these same performance these days as copying the value into a local variable, or just inlining the array access in an expression. This pattern was critical when we started writing the code circa 2013, however, since compilers then were very reluctant to issue vector instructions, and that's why it's peppered throughout the code.

tomidakn:
I agree that it would not (and theoretically should not, of course) affect the performance with modern compilers.
As far as @felker is happy, you can merge this commit.

forrestglines:
@felker I'll replace all the references with constant references on Monday,
it should be perfect to review then.

forrestglines:
@felker I've changed a few `Real&` to `const Real&`, plus I made constant a few `int`s and `Real`'s. Ready for review?

src/reconstruct/reconstruction.cpp
```diff
@@ -478,18 +476,18 @@ Reconstruction::Reconstruction(MeshBlock *pmb, ParameterInput *pin) :
...
 #pragma omp simd
         for (int i=(pmb->is)-NGHOST+1; i<=(pmb->ie)+NGHOST-1; ++i) {
-          Real& dx_im1 = pco->dx1f(i-1);
-          Real& dx_i   = pco->dx1f(i  );
-          Real& dx_ip1 = pco->dx1f(i+1);
-          Real qe = dx_i/(dx_im1 + dx_i + dx_ip1);       // Outermost coeff in CW eq 1.7
+          const Real& dx_im1 = pco->dx1f(i-1);
+          const Real& dx_i   = pco->dx1f(i  );
+          const Real& dx_ip1 = pco->dx1f(i+1);
+          const Real qe = dx_i/(dx_im1 + dx_i + dx_ip1);       // Outermost coeff in CW eq 1.7
```
felker, line 482:
```suggestion
          const Real qe = dx_i/(dx_im1 + dx_i + dx_ip1);  // Outermost coeff in CW eq 1.7
```

src/reconstruct/reconstruction.cpp
```diff
@@ -603,19 +601,19 @@ Reconstruction::Reconstruction(MeshBlock *pmb, ParameterInput *pin) :
...
 #pragma omp simd
           for (int j=(pmb->js)-NGHOST+2; j<=(pmb->je)+NGHOST-1; ++j) {
-            Real& dx_jm1 = pco->dx2f(j-1);
-            Real& dx_j   = pco->dx2f(j  );
-            Real& dx_jp1 = pco->dx2f(j+1);
-            Real qe = dx_j/(dx_jm1 + dx_j + dx_jp1);       // Outermost coeff in CW eq 1.7
+            const Real& dx_jm1 = pco->dx2f(j-1);
+            const Real& dx_j   = pco->dx2f(j  );
+            const Real& dx_jp1 = pco->dx2f(j+1);
+            const Real qe = dx_j/(dx_jm1 + dx_j + dx_jp1);       // Outermost coeff in CW eq 1.7
```
felker, line 607:
```suggestion
            const Real qe = dx_j/(dx_jm1 + dx_j + dx_jp1); // Outermost coeff in CW eq1.7
```

src/reconstruct/reconstruction.cpp
```diff
@@ -654,19 +652,19 @@ Reconstruction::Reconstruction(MeshBlock *pmb, ParameterInput *pin) :
...
 #pragma omp simd
         for (int k=(pmb->ks)-NGHOST+2; k<=(pmb->ke)+NGHOST-1; ++k) {
-          Real& dx_km1 = pco->dx3f(k-1);
-          Real& dx_k   = pco->dx3f(k  );
-          Real& dx_kp1 = pco->dx3f(k+1);
-          Real qe = dx_k/(dx_km1 + dx_k + dx_kp1);       // Outermost coeff in CW eq 1.7
+          const Real& dx_km1 = pco->dx3f(k-1);
+          const Real& dx_k   = pco->dx3f(k  );
+          const Real& dx_kp1 = pco->dx3f(k+1);
+          const Real qe = dx_k/(dx_km1 + dx_k + dx_kp1);       // Outermost coeff in CW eq 1.7
```
felker, line 658:
```suggestion
          const Real qe = dx_k/(dx_km1 + dx_k + dx_kp1);  // Outermost coeff in CW eq 1.7
```

felker:
Other files that likely have misused this (based on `grep -nri "real&" .`):
- `./hydro/new_blockdt.cpp`
- `./outputs/history.cpp`
- `./utils/matrix_mult.cpp`
- `./chemistry/cvode.cpp`
- a bunch of `nr_radiation/` files
- `./bvals/cc/nr_radiation/vacuum_rad.cpp`
- `./coordinates/cylindrical.cpp`
- `./pgen/chem_turb.cpp`
- `./pgen/strat.cpp`
- I should double check the EOS files, but I think they use it correctly

Issue #571, opened 2024-03-13
MHD shearing box has artifacts with SMR


c-white:
### Summary

The combination of MHD, SMR, and shear-periodic boundary conditions produces some artifacts and isn't fully supported right now. I propose the following, from short-term to long-term:

- [ ] Clarify in the documentation that this is not supported.
- [ ] Issue a runtime warning at least if this is detected.
- [ ] Figure out how to sync the flux corrections at the shearing boundary between different refinement levels.
  - If it makes the task easier, we might consider restricting refinement so that the shear boundary isn't too poorly behaved (perhaps by only allowing refinement variations in z there).

### Discussed in https://github.com/PrincetonUniversity/athena/discussions/569

<div type='discussions-op-text'>

<sup>Originally posted by **jonosquire** March  7, 2024</sup>
I have been trying to run an MHD shearing box with static mesh refinement and been running into issues. I add SMR in the vertical direction only for the usual shearing domain with shear_periodic boundary conditions in x. After a short time (~0.5 orbits), the integrator goes unstable with the timestep dropping to zero because of a very large radial magnetic field that develops in the 2-3 cells around the refinement location at the radial boundaries. Just 1 level of refinement causes the issue, but the blowup/instability occurs faster if higher levels are added. 

In order to diagnose the issue I have tried (in various combinations): the vl2, rk2, and rk3 time integrators with various conservative CFLs; order 2, 2c and 3 reconstruction; different initial conditions (from the strat.cpp pgen) and different field strengths; different orbital advection schemes (including no orbital advection); setting qshear=0 (still with the shear_periodic BCs); different mesh refinement locations, meshblock configurations and resolutions; different sound speeds and density floors. The problem persists with all of these. Turning off the magnetic field (pure hydro shearing box), or using periodic (rather than shear_periodic) boundary conditions in x1 removes the issue, and the grid setup appears to be as expected with all of the meshblocks at the same z facing the x1 shear-periodic boundaries having the same level.

Has anybody come across similar problems? I'm at a loss for what to try at this point – is there an issue with with shear_periodic BCs and SMR? I've attached an input file that should reproduce the issue with the strat.cpp pgen in the current athena version (this is effectively the default input file with different units, a finer grid, and the SMR added). I've also attached a screenshot showing the radial field blowup at the boundary (the colormap is saturated; the points right on the boundary at z=+-2.5 continue growing to infinity).

Thanks in advance for the help.
[...]

Issue #570, opened 2024-03-12
Leftover cosmic ray transport and diffusion changes


felker:
Follow up from #550 (mostly stuff for @yanfeij to generally weigh in on that PR)
- [ ] I still think that it is unfortunate that there is no consistency in variable names between the two CR algorithms. E.g. `Dpara/Dperp` vs. `sigma_diff(0:2, ...)`, initial condition set via `pcrdiff->ecr(k,j,i)` vs. `pcr->u_cr(4,k,j,i)`, even though they are in different modules and have different array dimensions as @tomidakn emphasizes
- [ ] https://github.com/PrincetonUniversity/athena/wiki/Cosmic-Ray-Transport only describes `cr/vmax` and `EnrollOpacityFunction()`. Need to describe some other parameters and functions in the class, especially those related to CR streaming and source functions. Also would be good to include the counterpart of the "equations solved" copied from Jiang and Oh (2018), assuming that they are different from the CR diffusion equations with the inclusion of the velocity-related terms. This is related to making the names consistent between CR diffusion and transport modules. Some items from `cr.hpp` header:

```
  Real vlim;
  Real max_opacity;
  AthenaArray<Real> sigma_adv;  // KGF: show an example of how this is set in the enroll functions 
...
  // Function in problem generators to update opacity
...
  void EnrollStreamingFunction(CRStreamingFunc MyStreamingFunction);

  void EnrollUserCRSource(CRSrcTermFunc my_func);
```

felker:
delaying until after the 24.0 release

Issue #553, opened 2023-12-28
Leftover chemistry fixes


felker:
I merged #496 since it was 99% complete thanks to @munan, happy new years everyone! GitHub's servers also seemed to be struggling with the long PR history and review comments, which were impossible to decipher given all the interleaving changes. I will document a few of the unaddressed comments here, which we can address now or eventually:
- @tomidakn 
  - `inputs/chemistry/athinput.chem_tigress` might be better excluded from the source code, since it requires TIGRESS data?
  - `src/utils/interp.cpp`  function names like `LP1D, LP2Di`, etc. are not self explanatory. Use longer names. (also, the Doxygen comment says " ID arrray linear interpolation", with typos at 1D and "array")
  - `src/units/units.cpp`, regarding things like ` code_time_cgs_` defined in this folder, "It is better to minimize problem specific codes in the master branch. Technically this can be done in the input file for each problem (i.e., always "custom"). Is there any reason to have them hard coded?"
- @felker  
  - `inputs/chemistry/athinput.pdr_moving` and other input files have `<problem>` block input variables like `Tw` which are seemingly unused in the problem generator file. If there is a reason for still keeping them in the input file, it should be documented in a comment

munan:
Thank you Kyle and Kengo!!! That's a great new year gift from you :)

I took notes of the issues and will fix them a bit later since they seem to
be minor.

felker:
delaying until after the 24.0 release

Issue #537, opened 2023-07-18
Leftovers from merge non-relativistic radiation transport: add more output switches, derive IMRadTaskList, etc.


felker:
Mostly from @tomidakn's review that we skipped over to finally merge the monster #492 pull request. 
- [ ] Derive `IMRadTaskList` from `TaskList` class to minimize code duplication (https://github.com/PrincetonUniversity/athena/pull/492#discussion_r1262260081)
- [ ] Add new , `rad` and `radmom` to `outputN/variable=` possible options e.g. (see https://github.com/PrincetonUniversity/athena/wiki/Outputs).
  - [ ] `cr` too? https://github.com/PrincetonUniversity/athena/pull/492/files/d68e6b8efd951c71e865c9f19a223baa2ece7c61#r1262189024
- [ ]  add some practical usage of the radiation transport and CR transport modules in the Wiki, including outputs and analyses. For example, Brief description of the test codes is helpful for users.
- [ ] Remove, rename, and/or cleanup `utils/gauleg.cpp`--- can we use or adapt `utils/gl_quadrature.cpp`?
>  The function is used when we construct angular grid in 1D for the radiation variable. It takes the user provided total number of angles and determine the values of \cos\theta for each angle and the associated weight for each angle based on GL quadrature. 
>
>  So 2*nmu is the total number of angles, wmu(i) is the weight for each angle and mu(0,0,0,n1,i) is the cos\theta value for each angle.

tomidakn:
Thank you again @yanfeij for all the efforts. I'm trying to use your CR module in a project with my student, and I noticed a few more things.

- CR Boundary conditions. Can we use user-defined physical BC for CR?
- User interface should be consistent with other modules. For example, currently the CR opacity is loaded through pcr->EnrollOpacityFunction in InitUserMeshBlockData, but for other modules we provide interfaces in Mesh. Retrospectively, we could move all the interfaces to each physics module and to be called in InitUserMeshBlockData, but anyway I believe we should provide coherently designed interface.

felker:
Related #502.

yanfeij:
> Thank you again @yanfeij for all the efforts. I'm trying to use your CR module in a project with my student, and I noticed a few more things.
> 
> * CR Boundary conditions. Can we use user-defined physical BC for CR?
[...]

PR #578, opened 2024-04-01
Release v24.0

felker:
Do we want to wait to merge any leftover fixes from chemistry, radiation+CR, and multigrid CR big pull requests (#553, #537, and #570, respectively)? @c-white do we wait for a fix for #571 ?


felker:
@swdavis, @tomidakn tells me that your group has a fix in the pipeline for a violation of the divergence conservation of mesh refinement in curvilinear coordinates? Is there a timeline on that, or should we push it to the `v24.1` release?

swdavis:
Yeah. I think my student (Maria Renee) planned to start a pull request today. She has forked the repo, implemented the fix, and tested it.  I went over the code with her quickly yesterday and addressed a few style issues.I think there is some slight tentativeness because there are some outstanding issues.  She implemented it only for 3D and there are some outstanding issues with using AMR with polar boundaries that are unresolved for spherical. There are some minor issues that she might want feedback from you and @tomidakn <https://github.com/tomidakn>, but I think those could be worked out quickly.  But, its up to you guys — not sure your timeline.

-Shane

felker:
ok sounds great! we can definitely hold the release and iterate on improving the PR--- just wanted to make sure that the timeline for submitting a draft wasnt indefinite

c-white:
I think we definitely don't want to wait for a full fix to #571. Maybe just a warning/error if a bad state is reached. Perhaps "if this block is at a shearing boundary and any of its neighbors have a different level, error."

For the record, I have never in my career run a shearing box, so maybe I'm not the best for making design decisions here.

felker:
Adding an error sounds good for now @c-white

PR #582, opened 2024-04-17
Divergence-preserving prolongation of coarse/fine shared face-centered fields for 3D spherical polar grids

mariareneemf:
## Description
 
I changed the `ProlongateSharedFieldX` functions in `mesh_refinement.cpp` to do the interpolation that defines the fine face-centered fields considering the areas of the overlapping fine and coarse cells for 3D spherical polar grids. 
To have access to the necessary coarse areas, I allocated memory for the `coord_area_x` arrays in the spherical polar file. These were previously skipped for coarse meshes with AMR. The coarse area arrays were defined in `mesh_refinement.hpp`.
I modified the field_loop_poles.cpp problem generator to include a divergence check of the magnetic field. This was used for testing and may not need to be part of a merge. I included the input file `athinput.field_loop_poles_checkdivb` used for this test.

The current AMR implementation does not preserve the divergence of the magnetic field during refinement in spherical polar grids.  Following a prescription suggested by @tomidakn, I have modified the interpolation that sets the fine shared face fields to enforce the following constraints:
 1. The transverse gradients on the fine cells match the (slope-limited) ones on the coarse cells.
 2. The total magnetic flux on the fine cells matches that on the coarse cell.
With these changes, the divergence is preserved during refinement.

- Interpolation in `ProlongateSharedFieldX` functions now preserves the divergence and magnetic flux during refinement of 3D spherical polar grids.
- Added divergence check to field loop poles problem generator and an input file to test it.

## Testing and validation

I ran the field loop poles test with AMR and checked the divergence of the magnetic field. To do this, I added a refinement condition in the problem generator to force refinement at a specific cycle in meshblocks with a magnetic field above a threshold and added a history output of the total divergence in the simulation. With the current prolongation scheme for shared face fields, the divergence increased by several orders of magnitude (~9) during refinement, as it does not preserve the magnetic flux. After the above-mentioned changes, the divergence change was less than an order of magnitude during refinement compared to the evolution of the divergence with no AMR. The divergence diagnostic was scaled with the number of cells used to compute it. This test can be run using the problem generator `field_loop_poles.cpp` with the input file `athinput.field_loop_poles_checkdivb`. Note that it is necessary to suppress the AMR-polar boundary error flag in `bvals/utils/check_polar.cpp` to run this test.

I ran the field loop poles test. I compiled the code with gcc 4.8.5, testing both, with openmpi 3.1.3 and no mpi.

The field loop poles test has a polar boundary, therefore, to run the field loop poles test it is necessary to comment L48-52 in `bvals/utils/check_polar.cpp` to suppress the AMR-polar boundary error. The test did not refine meshblocks adjacent to the pole so it was not a problem. I reverted this comment in `check_polar.cpp` as it cannot be generally applied.

## To-do

The changes are currently only implemented for 3D  but may be useful to implement in 2D and 1D and for other non-uniform coordinate systems. The changes should apply to any curvilinear coordinate system but are only triggered for spherical_polar coordinates since I have not tested them with other curvilinear coordinates.  Additionally, it may be useful to modify the polar boundary or refinement scheme to support AMR in spherical polar grids with polar boundaries. Finally, only the r coordinate used in the prolongation of shared face fields is area-averaged (L125 `src/coordinates/spherical_polar.cpp`). Face-area averaged coordinates could be also used for X2 and X3, although the difference may be very small.


buildbot-princeton:

felker:
ok to test

src/bvals/utils/check_polar.cpp
```diff
@@ -49,7 +49,7 @@ void BoundaryValues::CheckPolarBoundaries() {
       msg << "### FATAL ERROR in BoundaryValues constructor" << std::endl
           << "The use of AMR with any 'polar' or 'polar_wedge' boundary \n"
           << "flags is currently unsupported" << std::endl;
-      ATHENA_ERROR(msg);
+	  ATHENA_ERROR(msg);
```
felker, line 52:
```suggestion
      ATHENA_ERROR(msg);
```

src/coordinates/spherical_polar.cpp
```diff
@@ -144,8 +144,47 @@ SphericalPolar::SphericalPolar(MeshBlock *pmb, ParameterInput *pin, bool flag)
...
+    } else {
+      Real sm = std::abs(std::sin(x2f(jl  )));
+      Real sp = std::abs(std::sin(x2f(jl+1)));
+      Real cm = std::cos(x2f(jl  ));
+      Real cp = std::cos(x2f(jl+1));
+      coord_area1_j_(jl) = std::abs(cm - cp);
+      coord_area2_j_(jl) = sm;
+      coord_area2_j_(jl+1) = sp;
+    }
+  } else if (!coarse_flag) {
```
felker, line 187: change `else if` to `else` since this is a binary conditional

src/mesh/mesh_refinement.cpp
```diff
@@ -60,16 +60,24 @@ MeshRefinement::MeshRefinement(MeshBlock *pmb, ParameterInput *pin) :
         << "Reconfigure with an even number of ghost cells " << std::endl;
     ATHENA_ERROR(msg);
   }
-
+  int qx1;
   int nc1 = pmb->ncells1;
+  //In spherical polar grids, prolongation of shared face-centered fields requires face
```
felker, line 65: add space between `//` and text

felker:
Thanks for contributing! Still looking through it so I can understand the changes. 
> With the current prolongation scheme for shared face fields, the divergence increased by several orders of magnitude (~9) during refinement, as it does not preserve the magnetic flux. After the above-mentioned changes, the divergence change was less than an order of magnitude during refinement compared to the evolution of the divergence with no AMR

Shouldn't the divergence be zero within machine precision after this fix? Can you check/ report in absolute values vs. relative order of magnitudes?

I had completely forgotten that AMR (and also SMR?) + curvilinear coordinates causes the divergence constraint to be violated. A bit of an edge case, although the Stone (2020) method paper has such a test in Figure 31 for the MHD blast wave with AMR in spherical polar coordinates. The violation of the constraint is not mentioned in the paper, nor anywhere in the Wiki. We should document this and any expecations (and possibly emit a warning when MHD is enabled in the code in such cases?). 

Any though to turning your `inputs/mhd/athinput.field_loop_poles_checkdivb` and changes to `src/pgen/field_loop_poles.cpp` into a Python-based regression test?

src/mesh/mesh_refinement.cpp
```diff
@@ -787,6 +804,13 @@ void MeshRefinement::ProlongateSharedFieldX1(
         Real dx2p = x2p - x2c;
         const Real& fx2m = pco->x2s1(fj);
         const Real& fx2p = pco->x2s1(fj+1);
+        if (std::strcmp(COORDINATE_SYSTEM, "spherical_polar") == 0) {
```
tomidakn, line 807: While modern compilers should be able to optimize it automatically, can you move this if statement out of the loop for better optimization?

src/mesh/mesh_refinement.cpp
```diff
@@ -799,11 +823,31 @@ void MeshRefinement::ProlongateSharedFieldX1(
           Real gx3p = (coarse(k+1,j,i) - ccval)/dx3p;
           Real gx3c = 0.5*(SIGN(gx3m) + SIGN(gx3p))*std::min(std::abs(gx3m),
                                                              std::abs(gx3p));
-
-          fine(fk  ,fj  ,fi) = ccval - gx2c*(x2c - fx2m) - gx3c*(x3c - fx3m);
-          fine(fk  ,fj+1,fi) = ccval + gx2c*(fx2p - x2c) - gx3c*(x3c - fx3m);
-          fine(fk+1,fj  ,fi) = ccval - gx2c*(x2c - fx2m) + gx3c*(fx3p - x3c);
-          fine(fk+1,fj+1,fi) = ccval + gx2c*(fx2p - x2c) + gx3c*(fx3p - x3c);
+          if (std::strcmp(COORDINATE_SYSTEM, "spherical_polar") == 0) {
```
tomidakn, line 826: Same here

src/mesh/mesh_refinement.cpp
```diff
@@ -799,11 +823,31 @@ void MeshRefinement::ProlongateSharedFieldX1(
...
+            Real dfx2 = fx2p - fx2m;
+            Real dfx3 = fx3p - fx3m;
+            fine(fk  ,fj  ,fi) = ccval - gx2c*dfx2*(sarea_x1_[0][1](fi)
```
tomidakn, line 829: It is probably better (but not sure and can be compiler dependent) to load array quantities to local variables and reuse them as much as possible. For example,
```
Real carea = csarea_x1(i);
Real f00 = sarea_x1_[0][0](fi);
```
etc.

tomidakn: In fact, all the radial face areas are similar along the same (j,k) (even when the coordinate spacing is non uniform, so we can use the values for the first i and fi to save some computational cost, although I guess this part does not dominate the cost anyway.

src/mesh/mesh_refinement.cpp
```diff
@@ -881,11 +931,31 @@ void MeshRefinement::ProlongateSharedFieldX2(
...
                                                              std::abs(gx3p));
-
-          fine(fk  ,fj,fi  ) = ccval - gx1c*(x1c - fx1m) - gx3c*(x3c - fx3m);
-          fine(fk  ,fj,fi+1) = ccval + gx1c*(fx1p - x1c) - gx3c*(x3c - fx3m);
-          fine(fk+1,fj,fi  ) = ccval - gx1c*(x1c - fx1m) + gx3c*(fx3p - x3c);
-          fine(fk+1,fj,fi+1) = ccval + gx1c*(fx1p - x1c) + gx3c*(fx3p - x3c);
+          if (std::strcmp(COORDINATE_SYSTEM, "spherical_polar") == 0) {
+            Real dfx1 = fx1p - fx1m;
+            Real dfx3 = fx3p - fx3m;
+            fine(fk  ,fj  ,fi) = ccval - gx1c*dfx1*(sarea_x2_[0][0](fi+1)
```
tomidakn, line 937: This can cause 0/0 when the MeshBlock is touching the pole.
The easiest fix is just adding a small number to the denominator.

src/pgen/field_loop_poles.cpp
```diff
@@ -581,3 +605,105 @@ int RefinementCondition(MeshBlock *pmb) {
...
+          ncell += 1;
+        }
+      }
+    }
+  }
+  return ncell;
+}
+
+// Total number of cells in the simulation
+Real totcells(MeshBlock *pmb, int iout) {
```
tomidakn, line 697: You can use MeshBlock::GetNumberOfMeshBlockCells()

tomidakn (changes requested):
The 3D version looks good. We want to support 1D/2D and other coordinates (cylindrical in particular). I think we can just use the original code for Cartesian and the new code for other coordinate systems as the new one is general.

felker:
I accidentally committed your commits directly to the master branch on the command line when fixing some style things. I cannot easily revert the pseudo-merge nor re-open this PR. Can you open another PR containing @tomidakn's requested fixes? We can continue the discussion there. Sorry!

mariareneemf:
> Thanks for contributing! Still looking through it so I can understand the changes.
> 
> > With the current prolongation scheme for shared face fields, the divergence increased by several orders of magnitude (~9) during refinement, as it does not preserve the magnetic flux. After the above-mentioned changes, the divergence change was less than an order of magnitude during refinement compared to the evolution of the divergence with no AMR
> 
> Shouldn't the divergence be zero within machine precision after this fix? Can you check/ report in absolute values vs. relative order of magnitudes?
>  Yes, this plot illustrates it better. I am plotting the total sum of the absolute value of the divergence normalized by dividing by the magnitude of the magnetic field and multiplying by the minimum cell size. This quantity is then divided by the total number of cells that are used in computing it. The divergence diagnostic is preserved with the new changes.

flp_amr_divB

> Any though to turning your `inputs/mhd/athinput.field_loop_poles_checkdivb` and changes to `src/pgen/field_loop_poles.cpp` into a Python-based regression test?
I am not sure if it is optimal since the running of this test requires the suppression of the AMR-polar boundary error flagged in src/bvals/utils/check_polar.cpp. This error flag can't be generally removed since it could result in meshblocks adjacent to the pole having different refinement levels, which raises another error.

PR #584, opened 2024-04-26
Divergence-preserving prolongation of coarse/fine shared face-centered fields for 3D spherical polar and cylindrical grids 

mariareneemf:
## Description

The changes follow from pull request #582: Divergence-preserving prolongation of coarse/fine shared face-centered fields for 3D spherical polar grids. The divergence-preserving interpolation is now also triggered for 3D cylindrical grids. The condition for this is now evaluated outside the loops for performance. I also added area-averaged coordinates for spherical polar and cylindrical coordinates, which are used for MHD with AMR. Finally, I changed how the total cells in the field loop poles problem generator are computed: from a loop to using the function `MeshBlock::GetNumberOfMeshBlockCells()`.

The main changes were required to preserve the divergence of the magnetic field during refinement of curvilinear grids.

Interpolation in ProlongateSharedFieldX functions now preserves the divergence and magnetic flux during refinement of 3D spherical polar and cylindrical grids.

## Testing and validation

I ran the field loop poles test (using the input file `athinput.field_loop_poles_checkdivb`), forcing refinement at a specific cycle to check the divergence growth. The new prolongation scheme preserves the divergence.

 I compiled the code with gcc 4.8.5 and ran the field loop poles test.

The field loop poles test has a polar boundary, therefore, to run this test it is necessary to comment the AMR-polar boundary error flag in `bvals/utils/check_polar.cpp`. The test did not refine meshblocks adjacent to the pole so it was not a problem. I reverted this comment in `check_polar.cpp` as it cannot be generally applied.

### DIFF
--- a/src/mesh/mesh_refinement.cpp
+++ b/src/mesh/mesh_refinement.cpp
@@ -102,9 +102,9 @@ MeshRefinement::MeshRefinement(MeshBlock *pmb, ParameterInput *pin) :
   // Create coarse area arrays used in prolongation of shared face-centered fields in
   // curvilinear grids
   if (fluxinterp_) {
-    csarea_x1.NewAthenaArray(nc1);
-    csarea_x2.NewAthenaArray(nc1);
-    csarea_x3.NewAthenaArray(nc1);
+    csarea_x1_.NewAthenaArray(nc1);
+    csarea_x2_.NewAthenaArray(nc1);
+    csarea_x3_.NewAthenaArray(nc1);
   }
 
   // KGF: probably don't need to preallocate space for pointers in these vectors
@@ -819,11 +819,11 @@ void MeshRefinement::ProlongateSharedFieldX1(
           pco->Face1Area(fk,   fj+1, fsi, fei, sarea_x1_[0][1]);
           pco->Face1Area(fk+1, fj,   fsi, fei, sarea_x1_[1][0]);
           pco->Face1Area(fk+1, fj+1, fsi, fei, sarea_x1_[1][1]);
-          pcoarsec->Face1Area(k, j, si, ei, csarea_x1);
+          pcoarsec->Face1Area(k, j, si, ei, csarea_x1_);
           for (int i=si; i<=ei; i++) {
             int fi = (i - pmb->cis)*2 + pmb->is;
             Real ccval = coarse(k,j,i);
-            Real csarea = csarea_x1(i);
+            Real csarea = csarea_x1_(i);
             Real fsa00 = sarea_x1_[0][0](fi);
             Real fsa01 = sarea_x1_[0][1](fi);
             Real fsa10 = sarea_x1_[1][0](fi);
@@ -947,7 +947,7 @@ void MeshRefinement::ProlongateSharedFieldX2(
           int fj = (j - pmb->cjs)*2 + pmb->js;
           pco->Face2Area(fk,   fj,  fsi, fei, sarea_x2_[0][0]);
           pco->Face2Area(fk+1, fj,  fsi, fei, sarea_x2_[1][0]);
-          pcoarsec->Face2Area(k, j, si, ei, csarea_x2);
+          pcoarsec->Face2Area(k, j, si, ei, csarea_x2_);
           for (int i=si; i<=ei; i++) {
             int fi = (i - pmb->cis)*2 + pmb->is;
             const Real& x1m = pcoarsec->x1s2(i-1);
@@ -959,7 +959,7 @@ void MeshRefinement::ProlongateSharedFieldX2(
             const Real& fx1p = pco->x1s2(fi+1);
             Real ccval = coarse(k,j,i);
             Real dfx1 = fx1p - fx1m;
-            Real csarea = csarea_x2(i) + TINY_NUMBER;
+            Real csarea = csarea_x2_(i) + TINY_NUMBER;
             Real fsa00 = sarea_x2_[0][0](fi);
             Real fsa01 = sarea_x2_[0][0](fi+1);
             Real fsa10 = sarea_x2_[1][0](fi);
@@ -1091,7 +1091,7 @@ void MeshRefinement::ProlongateSharedFieldX3(
           Real dfx2 = fx2p - fx2m;
           pco->Face3Area(fk,   fj,  fsi, fei, sarea_x3_[0][0]);
           pco->Face3Area(fk, fj+1,  fsi, fei, sarea_x3_[0][1]);
-          pcoarsec->Face3Area(k, j, si, ei, csarea_x3);
+          pcoarsec->Face3Area(k, j, si, ei, csarea_x3_);
 
           for (int i=si; i<=ei; i++) {
             int fi = (i - pmb->cis)*2 + pmb->is;
@@ -1113,7 +1113,7 @@ void MeshRefinement::ProlongateSharedFieldX3(
             Real gx2c = 0.5*(SIGN(gx2m) + SIGN(gx2p))*std::min(std::abs(gx2m),
                                                              std::abs(gx2p));
             Real dfx1 = fx1p - fx1m;
-            Real csarea = csarea_x3(i);
+            Real csarea = csarea_x3_(i);
             Real fsa00 = sarea_x3_[0][0](fi);
             Real fsa01 = sarea_x3_[0][0](fi+1);
             Real fsa10 = sarea_x3_[0][1](fi);
--- a/src/mesh/mesh_refinement.cpp
+++ b/src/mesh/mesh_refinement.cpp
@@ -823,7 +823,7 @@ void MeshRefinement::ProlongateSharedFieldX1(
             Real dfx2 = fx2p - fx2m;
             Real dfx3 = fx3p - fx3m;
             Real ccval = coarse(k,j,i);
-	    Real csarea = csarea_x1(i);
+            Real csarea = csarea_x1(i);
             Real fsa00 = sarea_x1_[0][0](fi);
             Real fsa01 = sarea_x1_[0][1](fi);
             Real fsa10 = sarea_x1_[1][0](fi);
@@ -845,8 +845,8 @@ void MeshRefinement::ProlongateSharedFieldX1(
                                        + gx3c*dfx3*(fsa00 + fsa01)/csarea;
             fine(fk+1,fj+1,fi) = ccval + gx2c*dfx2*(fsa00 + fsa10)/csarea
                                        + gx3c*dfx3*(fsa00 + fsa01)/csarea;
-	  }
-	}
+          }
+        }
       }
     } else {
       for (int k=sk; k<=ek; k++) {
@@ -1020,8 +1020,8 @@ void MeshRefinement::ProlongateSharedFieldX2(
             fine(fk  ,fj,fi+1) = ccval + gx1c*(fx1p - x1c) - gx3c*(x3c - fx3m);
             fine(fk+1,fj,fi  ) = ccval - gx1c*(x1c - fx1m) + gx3c*(fx3p - x3c);
             fine(fk+1,fj,fi+1) = ccval + gx1c*(fx1p - x1c) + gx3c*(fx3p - x3c);
-	  }
-	}
+          }
+        }
       }
     }
   } else if (pmb->block_size.nx2 > 1) {
@@ -1164,8 +1164,8 @@ void MeshRefinement::ProlongateSharedFieldX3(
             fine(fk,fj  ,fi+1) = ccval + gx1c*(fx1p - x1c) - gx2c*(x2c - fx2m);
             fine(fk,fj+1,fi  ) = ccval - gx1c*(x1c - fx1m) + gx2c*(fx2p - x2c);
             fine(fk,fj+1,fi+1) = ccval + gx1c*(fx1p - x1c) + gx2c*(fx2p - x2c);
-	  }
-	}
+          }
+        }
       }
     }
   } else if (pmb->block_size.nx2 > 1) {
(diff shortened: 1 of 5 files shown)


buildbot-princeton:

tomidakn (changes requested):
The code looks good, just a minor style issue.

tomidakn:
Hi @mariareneemf , thank you for the commit!
The code looks good to me. Can you post your a figure of the test result with the latest code here?

src/mesh/mesh_refinement.cpp
```diff
@@ -786,63 +794,91 @@ void MeshRefinement::ProlongateSharedFieldX1(
...
           pco->Face1Area(fk+1, fj,   fsi, fei, sarea_x1_[1][0]);
           pco->Face1Area(fk+1, fj+1, fsi, fei, sarea_x1_[1][1]);
           pcoarsec->Face1Area(k, j, si, ei, csarea_x1);
-        }
-        for (int i=si; i<=ei; i++) {
-          int fi = (i - pmb->cis)*2 + pmb->is;
-          Real ccval = coarse(k,j,i);
+          for (int i=si; i<=ei; i++) {
+            int fi = (i - pmb->cis)*2 + pmb->is;
+            Real dfx2 = fx2p - fx2m;
```
tomidakn, line 823: You can move this line out of the i-loop  (to after l.815)

src/mesh/mesh_refinement.cpp
```diff
@@ -786,63 +794,91 @@ void MeshRefinement::ProlongateSharedFieldX1(
...
+            Real dfx3 = fx3p - fx3m;
```
tomidakn, line 824: And this out of the j-loop to l.807.

src/mesh/mesh_refinement.cpp
```diff
@@ -896,67 +932,96 @@ void MeshRefinement::ProlongateSharedFieldX2(
...
+            const Real& x1m = pcoarsec->x1s2(i-1);
+            const Real& x1c = pcoarsec->x1s2(i);
+            const Real& x1p = pcoarsec->x1s2(i+1);
+            Real dx1m = x1c - x1m;
+            Real dx1p = x1p - x1c;
+            const Real& fx1m = pco->x1s2(fi);
+            const Real& fx1p = pco->x1s2(fi+1);
+            Real ccval = coarse(k,j,i);
             Real dfx1 = fx1p - fx1m;
             Real dfx3 = fx3p - fx3m;
```
tomidakn, line 961: dfx3 does not depends on i and j, so this line can be moved out of the i,j loops.

src/mesh/mesh_refinement.cpp
```diff
@@ -1011,67 +1076,96 @@ void MeshRefinement::ProlongateSharedFieldX3(
...
-          Real gx2m = (ccval - coarse(k,j-1,i))/dx2m;
-          Real gx2p = (coarse(k,j+1,i) - ccval)/dx2p;
-          Real gx2c = 0.5*(SIGN(gx2m) + SIGN(gx2p))*std::min(std::abs(gx2m),
+            Real gx2m = (ccval - coarse(k,j-1,i))/dx2m;
+            Real gx2p = (coarse(k,j+1,i) - ccval)/dx2p;
+            Real gx2c = 0.5*(SIGN(gx2m) + SIGN(gx2p))*std::min(std::abs(gx2m),
                                                              std::abs(gx2p));
-          if (std::strcmp(COORDINATE_SYSTEM, "spherical_polar") == 0) {
             Real dfx1 = fx1p - fx1m;
             Real dfx2 = fx2p - fx2m;
```
tomidakn, line 1115: same here

tomidakn (changes requested):
Sorry, I noticed a few more points (all minor, and probably very little impact on performance)

felker:
ok to test

felker:
@mariareneemf what editor do you use? I would suggest flipping the settings to turn TAB characters to spaces and to delete trailing whitespace whenever you save the file. Thanks for the fixes!

mariareneemf:
> Hi @mariareneemf , thank you for the commit! The code looks good to me. Can you post your a figure of the test result with the latest code here?

Hi @tomidakn, sure!
divB_fieldloopoles

felker:
good choice of editor 😄 

```lisp
;; Automatically remove trailing whitespace when file is saved.
(add-hook 'c-mode-common-hook
      (lambda()
        (add-hook 'write-contents-functions
              (lambda()
                 (save-excursion
                   (delete-trailing-whitespace))))))
(defconst athena-ncols 90)
(defun athena-cpp-mode ()
  (setq fill-column athena-ncols)
  (setq indent-tabs-mode nil)
  ;; https://google.github.io/styleguide/cppguide.html#Spaces_vs._Tabs
  (setq tab-width 2))
;; Use "M-x whitespace-mode" to show spaces as dots, tabs as >>
;; Use "M-x untabify" or "M-x tabify" to convert between the two indentations
(add-hook 'c-mode-common-hook #'athena-cpp-mode)
(add-hook 'c-mode-common-hook #'turn-on-auto-fill)

;; Add line numbers to the left column
(add-hook 'c-mode-common-hook #'display-line-numbers-mode)

```

I also recommend the `google-c-style` package for Athena++ soruce

mariareneemf:
> good choice of editor 😄
> 
> ```commonlisp
> ;; Automatically remove trailing whitespace when file is saved.
> (add-hook 'c-mode-common-hook
>       (lambda()
>         (add-hook 'write-contents-functions
>               (lambda()
>                  (save-excursion
>                    (delete-trailing-whitespace))))))
> (defconst athena-ncols 90)
> (defun athena-cpp-mode ()
>   (setq fill-column athena-ncols)
>   (setq indent-tabs-mode nil)
>   ;; https://google.github.io/styleguide/cppguide.html#Spaces_vs._Tabs
>   (setq tab-width 2))
> ;; Use "M-x whitespace-mode" to show spaces as dots, tabs as >>
> ;; Use "M-x untabify" or "M-x tabify" to convert between the two indentations
> (add-hook 'c-mode-common-hook #'athena-cpp-mode)
> (add-hook 'c-mode-common-hook #'turn-on-auto-fill)
> 
> ;; Add line numbers to the left column
> (add-hook 'c-mode-common-hook #'display-line-numbers-mode)
> ```
> 
> I also recommend the `google-c-style` package for Athena++ soruce

Thanks!

tomidakn (approved):

tomidakn:
I think this is good.

One thing to note is other gr-related coordinate systems. 
I guess it is safe to use the new prolongation for everything other than Cartesian/Minkowski.
How do you think, @c-white ?

Also, this is minor but even in curvilinear coordinates, some surfaces do not need the new scheme.
For example, X1/X3 surfaces in cylindrical coordinates have no problem with the original formula.
But this probably has only minor impact on performance, and anyway the new formula is more general.

felker:
I dont think the Wiki was ever updated--- should we add something to the following pages: AMR, SMR, Coordinate Systems?

smressle:
This is great!  I am currently running into this issue for AMR with a general metric in GRMHD, to follow up on @tomidakn's comment.  So I may try to implement something similar in gr_user.

tomidakn:
@smressle that is great. Please let us know the outcome!

smressle:
OK so I was able to quickly add the new algorithm to the ProlongateShared functions and was still seeing fairly high divergences.  I think the Restriction algorithm also might need to be changed, e.g.

` Real tarea = sarea_x1_[0][0](i) + sarea_x1_[0][1](i) +
                       sarea_x1_[1][0](i) + sarea_x1_[1][1](i);`
                       
Should maybe be something like

` Real tarea = csarea_x1_(ci);`

Because the areas of the four fine faces might not add up to the coarse area.

But after doing that I still have issues.  So I am thinking that ProlongateInternalField might have to be altered?  I am trying to understand the interpolation algorithm right now so I am not sure how it might need to be changed.  Any thoughts?

smressle:
Ah, I think ProlongateInternalField should be fine as long as ProlongateShared does it's job in preserving div B because ProlongateInternal Field only operates on the previously set fine fields on shared faces, right?

smressle:
Hi @mariareneemf,  thanks for all your work on this!  Yes I need to define the area averaged coordinates, will do.  And thanks for the clarification on the ProlongateInternalField function. 

I am using this for divb:

'  for(int k=ks; k<=ke; k++) {
    for(int j=js; j<=je; j++) {
      pmb->pcoord->Face1Area(k,   j,   is, ie+1, face1);
      pmb->pcoord->Face2Area(k,   j+1, is, ie,   face2p);
      pmb->pcoord->Face2Area(k,   j,   is, ie,   face2m);
      pmb->pcoord->Face3Area(k+1, j,   is, ie,   face3p);
      pmb->pcoord->Face3Area(k,   j,   is, ie,   face3m);
      for(int i=is; i<=ie; i++) {
        divb+=(face1(i+1)*b.x1f(k,j,i+1)-face1(i)*b.x1f(k,j,i)
              +face2p(i)*b.x2f(k,j+1,i)-face2m(i)*b.x2f(k,j,i)
              +face3p(i)*b.x3f(k+1,j,i)-face3m(i)*b.x3f(k,j,i));
      }
    }
  }
'

But I do think I found a gr-specific problem in ProlongateInternalField:
'          Real Sdx1 = SQR(pco->dx1f(fi) + pco->dx1f(fi+1));
          Real Sdx2 = SQR(pco->GetEdge2Length(fk+1,fj,fi+1) +
                          pco->GetEdge2Length(fk+1,fj+1,fi+1));
          Real Sdx3 = SQR(pco->GetEdge3Length(fk,fj+1,fi+1) +
                          pco->GetEdge3Length(fk+1,fj+1,fi+1));
          Uxx *= 0.125; Vyy *= 0.125; Wzz *= 0.125;
          Uxyz *= 0.125/(Sdx2 + Sdx3);
          Vxyz *= 0.125/(Sdx1 + Sdx3);
          Wxyz *= 0.125/(Sdx1 + Sdx2);'
          
 I don't see why the Edgelength is not used for the x1 coordinate.  For cartesian, spherical polar, and cylindrical it's the same thing.  But not for GR where there is a factor of sqrt(-g) difference.  In fact, in general Sdx1 would not necessarily have the same units as the other two Sdx2 and Sdx3.  
 
 And perhaps the sum of the Edgelengths should be replaced by the coarse edge length?

mariareneemf:
I think for the divergence one should consider the absolute value, as the divergence is preserved for each cell and you do not want the positive divergence of one cell possibly canceling out with the negative divergence of another in the total sum. I am also normalizing it with the cell-centered field and a length scale, as the magnetic energy may grow and the lengths of the cells can also change. 

I agree that EdgeLength should also be used for Sdx1. The EdgeLengths used there are from the fine cells according to the Toth & Roe (2002) method paper. I am not sure where the sum of the two fine EdgeLengths comes from though.

bitwalkabout:
Hi all! My purview on this issue is still comparatively limited, but I might have found something. Last week I realized I also had B-field divergence errors (running with a gr_user setup in sph-polar coordinates). Friday I narrowed this down to specifically during prolongation and so I haven't had much time to look through the code. I was excited and relieved to find ya'll were already working on the issue but implementing the newest pushed changes to master branch wasn't sufficient to solve the issue in the general spacetime setup I've got.

However, I believe I have got it working with the following hack:   Within ProlongateSharedFieldX[1,2,3] functions, I found that the net field over the combined fine faces was off by a small amount after interpolation adjustments. The following solved this (for x1 direction, similarly done for the other two in 3D):
*********************
Real rreset = (ccval*csarea)/(fine(fk,fj,fi)*fsa00 + fine(fk,fj+1,fi)*fsa01+ fine(fk+1,fj,fi)*fsa10+ fine(fk+1,fj+1,fi)*fsa11) ;                                                  
fine(fk,fj,fi) *=rreset ;                                                                                                                                                         
fine(fk,fj+1,fi) *=rreset ;
fine(fk+1,fj,fi) *=rreset ;
fine(fk+1,fj+1,fi) *=rreset ;
*********************
Given the discussion in the AMR paper I was under the impression that this sort of renormalization was already done, but perhaps that functionality is not being accessed somehow (I was not able to find the associated renormalizing code) in my setup, for gr_user, or for some other reason. 

 I would appreciate assistance in implementing this change in a clean way, though, since I'm still very new to the AMR routines in the code. I'm sure there are issues with my hack but it does seem to fix the CT violations for me.

c-white:
I agree that any appearances of `GetEdge2Length` and `GetEdge3Length` should be accompanied by `GetEdge1Length` in GR.

Also, it's worth clarifying the difference between built-in GR coordinates (Schwarzschild, spherical Kerr-Schild) and user-defined coordinates. In the former, the code always evaluates integrals of metric factors over cells, faces, and edges exactly. As a result, the sum of the fine volumes/areas/lengths is exactly the same as the corresponding coarse volume/area/length. Thus no corrections are needed, just as they are not needed in Newtonian coordinates.

With user-defined coordinates this is no longer feasible, so instead the code approximates such integrals by pulling $\sqrt{-g}$ out of the integrand, evaluated at some nominal central point. This will, as observed, have an adverse effect on exact preservation of certain quantities across refinement boundaries.

smressle:
Alright so I believe I have a working implementation that preserved divB in arbitrary GRUser coordinates.  The functions that need to be modified are:

RestrictFieldX1, RestrictFieldX2, RestrictFieldX3

ProlongateSharedFieldX1, ProlongateSharedFieldX2, ProlongateSharedFieldX3

and 

FaceCenteredBoundaryVariable::LoadFluxBoundaryBufferToCoarser

The latter function needs to take into account that edgelengths do not add up exactly from fine to coarse and so operate directly on E * length instead of just E.  Otherwise the emfs on neighboring coarse/fine blocks are inconsistent so that they will store different magnetic fields for the same faces.  This introduces divB when there is a restriction to a coarser grid at one of these locations. I can post the code and later add to the latest version.

smressle:
Also, it is important to note that the same issues that affect divB conservation affect hydro variable conservation.  So the similar functions for the cell centered variables need to be modified in similar ways in order to ensure full energy, momentum, and mass conservation.

PR #585, opened 2024-05-07
Add exclusion for SMR+MHD+shearing box

felker:
See #571 and #569

PR #591, opened 2024-06-06
Add better exclusion for MHD+shearing box+SMR with refinement in x3 dir at shearing boundary

felker:
Alternative to #585 suggested by @tomidakn that might be more narrowly-tailored. 
See https://github.com/PrincetonUniversity/athena/issues/571 and https://github.com/PrincetonUniversity/athena/discussions/569.

Ideally, we would want to check the refinement restrictions only at the beginning of the simulation and after refinement (if adaptive)-- but I am not really sure I implemented this check correctly. @tomidakn am I correct in interpreting `nblx3 = pmy_mesh_->nrbx3*(1L << level);` as  "the maximum possible number of MeshBlocks in the x3 direction across the entire mesh at the same level of refinement as the current MeshBlock"? So that the following traverses the root `MeshBlockTree` to check that every MeshBlock at the same (x1, x2) logical location as the current MeshBlock (which is on a shearing domain boundary) is on the same refinement level?
```
          for (int64_t lx3=0; lx3<nblx3; lx3++) {
            loc.lx3   = lx3;
            MeshBlockTree *mbt = pmy_mesh_->tree.FindMeshBlock(loc);
...
            if ((mbt == nullptr || mbt->GetGid() == -1) && MAGNETIC_FIELDS_ENABLED) {
...
```

Along with the existing similar check for the blocks at the same (x1, x3) logical location, but at all other x2, at the shearing boundary, this should ensure that at any timestep all blocks on the shearing boundaries are at the same level of refinement, I think? Note, this existing check along x2 is run for both Hydro and MHD, since I assume the refinement along the shearing direction breaks even the hydro flux corrections. 

Since these checks are in the BoundaryValues constructor, I think it also holds during initialization? I havent really thought about the 2D `shboxcoord=2` case, but I think this is sufficient

Questions for @tomo-ono on shearing box stuff:
- Is `shboxcoord=2` "2D x-z case" only ever used if `mesh/nx3=1`, and similarly `shboxcoord=1` only in 3D simulations? If so, isnt the switch redundant?
- [ ] Can we add a 2D shearing sheet diagram to https://github.com/PrincetonUniversity/athena/wiki/Shearing-Box alongside the conventional 3D shearing box diagram that is there? It would have helped me when reading and editing the shear code. It especially gets confusing when dealing with `shboxcoord=2`  case and thinking in (x1,x2)=(x,z) with y into the page?

- [ ] Remove the error if AMR is used with shearing box; leave it up to the user to define a smart `RefinementCondition()` that guarantees that the requirements arent violated? Keep a runtime warning?
https://github.com/PrincetonUniversity/athena/blob/f701e6785d710ef6a90fb23254e4932b8de27ff6/src/bvals/bvals.cpp#L273-L278

### DIFF
--- a/src/bvals/bvals.cpp
+++ b/src/bvals/bvals.cpp
@@ -60,6 +60,7 @@
 BoundaryValues::BoundaryValues(MeshBlock *pmb, BoundaryFlag *input_bcs,
                                ParameterInput *pin)
     : BoundaryBase(pmb->pmy_mesh, pmb->loc, pmb->block_size, input_bcs), pmy_block_(pmb),
+      is_shear{}, loc_shear{0, pmy_mesh_->nrbx1*(1L << pmb->loc.level - 1)},
       sb_data_{}, sb_flux_data_{} {
   // Check BC functions for each of the 6 boundaries in turn ---------------------
   for (int i=0; i<6; i++) {
@@ -122,13 +123,6 @@ BoundaryValues::BoundaryValues(MeshBlock *pmb, BoundaryFlag *input_bcs,
   // set parameters for shearing box bc and allocate buffers
   shearing_box = 0;
   if (pmy_mesh_->shear_periodic) {
-    // see discussion #569
-    if (pmy_mesh_->multilevel && MAGNETIC_FIELDS_ENABLED) {
-      std::stringstream msg;
-      msg << "### FATAL ERROR in BoundaryValues Class" << std::endl
-          << "SMR is incompatible with shearing box with MHD" << std::endl;
-      ATHENA_ERROR(msg);
-    }
     // It is required to know the reconstruction scheme before pmb->precon is defined.
     // If a higher-order scheme than PPM is implemented, xgh_ must be larger than 2.
     std::string input_recon = pin->GetOrAddString("time", "xorder", "2");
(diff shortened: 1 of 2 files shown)


tomo-ono:
@felker 
>Is shboxcoord=2 "2D x-z case" only ever used if mesh/nx3=1, and similarly shboxcoord=1 only in 3D simulations? If so, isnt the switch redundant?
Yes, "shboxcoord=2" is only used in the 2D x-z case. I understand that it is redundant. However, I don't have better idea.
> Can we add a 2D shearing sheet diagram to https://github.com/PrincetonUniversity/athena/wiki/Shearing-Box alongside the conventional 3D shearing box diagram that is there? It would have helped me when reading and editing the shear code. It especially gets confusing when dealing with shboxcoord=2 case and thinking in (x1,x2)=(x,z) with y into the page?
I agree with you. A figure for the 2D x-z case must be helpful. I just reuse a 3D figure which exists before I implemented it. 
>Remove the error if AMR is used with shearing box; leave it up to the user to define a smart RefinementCondition() that guarantees that the requirements arent violated? Keep a runtime warning?
This is very difficult question. In my opinion, I would like to keep the conservative error. However, we should give a path to pass it for users who knows the risk.

felker:
retest this please

felker:
retest this please

tomidakn:
> @tomidakn am I correct in interpreting nblx3 = pmy_mesh_->nrbx3*(1L << level); as "the maximum possible number of MeshBlocks in the x3 direction across the entire mesh at the same level of refinement as the current MeshBlock"? So that the following traverses the root MeshBlockTree to check that every MeshBlock at the same (x1, x2) logical location as the current MeshBlock (which is on a shearing domain boundary) is on the same refinement level?

Yes, this is correct.

I think this check should work. On one hand, this is a bit redundant as it checks same MeshBlocks many times, but on the other hand it is *not* redundant as it is parallelized and only checks MeshBlocks owned by one rank. I think it is fine.

tomidakn (approved):
I think this is good, but I would appreciate if @tomo-ono could double check.

tomo-ono:
I checked the changes. I think that these changes are good.

Issue #617, opened 2024-08-29
inter-block divergence issue when refinining under certain circumstances, generic to all coordinates


bitwalkabout:
The issue involves large magnetic divergence measured in the first row/plane of ghost zones outside the active region at an inter-mesh boundary. This issue arrises in a very specific circumstance: if left meshblock A is higher resolution than right meshblock B, and the refinement boundary moves to the right such that meshblock B matches the resolution of A, the first ghost zones have a large divergence, capable of affecting evolution. This seems to be true for all three dimensions in 3D, but only when the coarse meshblock side is refined, not when the fine meshblock is restricted.

I first came across this bug because of off-disk issues running the native Kerr-Schild coordinates with AMR turned on (forced to be axisymmetric). The recent divergence fixes related to curvilinear coordinates fixed other instances of large divergence -- namely, inside the active domain -- but I later saw hydrodynamic discontinuities that appeared like divergence-led errors so I checked inside the ghost zones and found the divergence there.

I hypothesized a cause that would be generic to flat spacetime, cartesian coordinates, so turned attention to the rayleigh taylor 3d problem with AMR.

To replicate, consider commit 1591aab84ba7055e5b356a8f069695ea451af8a0, running the RT with the following configuration:
configure.py -b --prob rt --flux hlle --nghost 4 -mpi -hdf5 -h5double

I change the refinement condition in rt.cpp to:
-  if (drmax > 1.5) return 1;
-  else if (drmax < 1.2) return -1;
+  if (drmax > 1.1) return 1;
+  else if (drmax < 1.05) return -1;

and, rather than looking at every meshblock output individually, to get a rough look at the ghost zone divergence, I calculate the divergence for all active cells -3 (shifting left), and write this to a user defined output variable. This is the only other change to rt.cpp.

I use athinput.rt3d with only a few changes:
> I output to hdf5
> xorder ->3
> for speed  I reduced resolution to
nx1 = 36, nx2=6, nx3=36
> I turned on mesh refinement:
refinement = adaptive
numlevel       = 3
derefine_count    = 5

<meshblock>
nx1 = 6  
nx2 = 6  
nx3 = 6  

> also changed to: 
iprob = 3
amp   = 0.5
drat  = 1.4
b0 = .0001
angle = 30.

The Kerr-Schild issue appears as divergence lines:
![image](https://github.com/user-attachments/assets/4a5f2b95-27c4-4cf5-b526-8cc387f92d0c)

[...]

PR #625, opened 2024-10-12
Fixing divergence error with AMR (Issue #617)

tomidakn:
This is to fix the divergence error with AMR reported in #617.

When a MeshBlock contacting a finer MeshBlock is refined, the original code refined normal B-fields on the shared face by interpolating the coarser fields, but they do not exactly match the existing finer fields on the already refined neighbor MeshBlock. With this commit, the shared face fields are copied from the finer MeshBlock, hence enforcing the divergence B constraint.

## Prerequisite checklist
- [X] My code follows the Athena++ [Style Guide](https://github.com/PrincetonUniversity/athena/wiki/Style-Guide)
- [X] All new and existing tests passed.

### DIFF
--- a/src/athena.hpp
+++ b/src/athena.hpp
@@ -78,6 +78,26 @@ struct LogicalLocation { // aggregate and POD type
 //! prototype for overloading the comparison operator (defined in meshblock_tree.cpp)
 bool operator==(const LogicalLocation &l1, const LogicalLocation &l2);
 
+
+//! \fn inline std::int64_t rotl(std::int64_t i, int s)
+//  \brief left bit rotation function for 64bit integers (unsafe if s > 64)
+
+inline std::int64_t rotl(std::int64_t i, int s) {
+  return (i << s) | (i >> (64 - s));
+}
+
+
+//! \struct LogicalLocationHash
+//  \brief Hash function object for LogicalLocation
+
+struct LogicalLocationHash {
+ public:
+  std::size_t operator()(const LogicalLocation &l) const {
+    return static_cast<std::size_t>(l.lx1^rotl(l.lx2,21)^rotl(l.lx3,42));
+  }
+};
+
+
 //----------------------------------------------------------------------------------------
 //! \struct RegionSize
 //! \brief physical size and number of cells in a Mesh or a MeshBlock
--- a/src/mesh/amr_loadbalance.cpp
+++ b/src/mesh/amr_loadbalance.cpp
@@ -1225,7 +1225,7 @@ void Mesh::ProlongateMeshBlock(MeshBlock *pb) {
 
   int il = pb->cis, iu = pb->cie+1, jl = pb->cjs, ju = pb->cje + f2,
       kl = pb->cks, ku = pb->cke + f3;
-  // Step FFC6. skip the surface fields contacting previously refined cells
+  // Step FFC8. skip the surface fields contacting previously refined MeshBlocks
   if (pmr->flag_ffc_recv_[BoundaryFace::inner_x1]) il++;
   if (pmr->flag_ffc_recv_[BoundaryFace::outer_x1]) iu--;
   if (pmr->flag_ffc_recv_[BoundaryFace::inner_x2]) jl++;
@@ -1272,7 +1272,7 @@ void Mesh::PrepareAndSendFaceFieldCorrection(LogicalLocation *newloc,
     }
   }
 
-  // Step FFC1. Find pairs that need face field correction
+  // Step FFC2. Find pairs that need face field correction
   int s = 0;
   for (int n : refined_) { // loop over refined MeshBlocks
     LogicalLocation const &nloc = newloc[n];
@@ -1375,7 +1375,7 @@ void Mesh::PrepareAndSendFaceFieldCorrection(LogicalLocation *newloc,
     }
   }
 
-  // Step FFC2. Initiate MPI receive
+  // Step FFC3. Initiate MPI receive
 #ifdef MPI_PARALLEL
   for (FaceFieldCorrection &t : ffc_recv_) {
     if (ranklist[t.from] != Globals::my_rank) {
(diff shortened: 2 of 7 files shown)


changgoo:
failure from the `pgen_compile` test:
```
src/mesh/amr_loadbalance.cpp:1223:47: error: implicit conversion loses integer precision: 'std::vector::size_type' (aka 'unsigned long') to 'const int' [-Werror,-Wshorten-64-to-32]
  const int nf = my_blocks(0)->pmr->pvars_fc_.size();
            ~~   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
1 error generated.
```

src/mesh/amr_loadbalance.cpp
```diff
@@ -1194,22 +1154,310 @@ void Mesh::FinishRecvCoarseToFineAMR(MeshBlock *pb, Real *recvbuf) {
...
+//! \fn void Mesh::PrepareAndSendFaceFieldCorrection(LogicalLocation *newloc,
+//!                              int *ranklist, int *newrank, int *nslist, int nbtold)
+//! \brief the first part of the face field correction
+
+void Mesh::PrepareAndSendFaceFieldCorrection(LogicalLocation *newloc,
+                                  int *ranklist, int *newrank, int *nslist, int nbtold) {
+  const int bnx1 = my_blocks(0)->block_size.nx1;
+  const int bnx2 = my_blocks(0)->block_size.nx2;
+  const int bnx3 = my_blocks(0)->block_size.nx3;
+  const int nf = my_blocks(0)->pmr->pvars_fc_.size();
```
felker, line 1223:
```suggestion
  const int nf = static_cast<int>(my_blocks(0)->pmr->pvars_fc_.size());
```

tomidakn: I agree with this suggestion.

changgoo:
another error:
```
src/mesh/amr_loadbalance.cpp: In member function ‘void Mesh::PrepareAndSendFaceFieldCorrection(LogicalLocation*, int*, int*, int*, int)’:
src/mesh/amr_loadbalance.cpp:1296:48: error: suggest parentheses around arithmetic in operand of ‘^’ [-Werror=parentheses]
 1296 |             ffc_send_.emplace_back(i, n, xf2^1 + 2, sffc2, -1);
      |                                              ~~^~~
src/mesh/amr_loadbalance.cpp:1326:48: error: suggest parentheses around arithmetic in operand of ‘^’ [-Werror=parentheses]
 1326 |             ffc_send_.emplace_back(i, n, xf3^1 + 4, sffc3, -1);
      |                                              ~~^~~
```

src/mesh/amr_loadbalance.cpp
```diff
@@ -1194,22 +1154,310 @@ void Mesh::FinishRecvCoarseToFineAMR(MeshBlock *pb, Real *recvbuf) {
...
+        else
+          fskip = true;
+      }
+      if (!fskip) {
+        auto itr = locmap_[lev].find(floc);
+        if (itr != locmap_[lev].end()) {
+          const auto& it = *itr;
+          int i = it.second;
+          if (ranklist[i] == Globals::my_rank) {
+            ffc_send_.emplace_back(i, n, xf2^1 + 2, sffc2, -1);
```
felker, line 1296:
```suggestion
            ffc_send_.emplace_back(i, n, xf2^(1 + 2), sffc2, -1);
```

src/mesh/amr_loadbalance.cpp
```diff
@@ -1194,22 +1154,310 @@ void Mesh::FinishRecvCoarseToFineAMR(MeshBlock *pb, Real *recvbuf) {
...
+        else
+          fskip = true;
+      }
+      if (!fskip) {
+        auto itr = locmap_[lev].find(floc);
+        if (itr != locmap_[lev].end()) {
+          const auto& it = *itr;
+          int i = it.second;
+          if (ranklist[i] == Globals::my_rank) {
+            ffc_send_.emplace_back(i, n, xf3^1 + 4, sffc3, -1);
```
felker, line 1326:
```suggestion
            ffc_send_.emplace_back(i, n, xf3^(1 + 4), sffc3, -1);
```

felker:
@tomidakn can you check that my latest suggestion is what is intendend? I had to remind myself of the counter-intuitivie operator precedence rules in C re: bitwise operators and arithmetic operators. Interestingly, ChatGPT 4o-mini gets it wrong: 
image

src/mesh/amr_loadbalance.cpp
tomidakn: Oh wait... I made the same mistake as Chat-GPT. It should be (xf3^1)+4. How could it be working...!?

tomidakn:
I'm afraid I made the same mistake as Chat-GPT.

src/mesh/amr_loadbalance.cpp
felker: would be curious to hear if you come up with an explanation of how it worked for you locally before... were you running only a few tests?

felker:
don't worry, I have scolded the AI
image
image

src/mesh/amr_loadbalance.cpp
tomidakn: This error affects only ghost cells in the x2 and x3 surfaces, and I was checking only the x1 surface. Another lesson that covering all the test cases is important.

changgoo:
retest this on stellar

felker:
> Wow, I also didn't appreciate the pgen_compile test much and often blame that it is too picky. But, now I see why we have that!

see, I didnt just implement these strict tests just to be annoying! 😂

src/athena.hpp
```diff
@@ -78,6 +78,26 @@ struct LogicalLocation { // aggregate and POD type
 //! prototype for overloading the comparison operator (defined in meshblock_tree.cpp)
 bool operator==(const LogicalLocation &l1, const LogicalLocation &l2);
 
+
```
felker, line 81: Could you say a bit about why this moved from being a (mostly?) "multigrid-specific" feature?
I see that the `LogicalLocationHash` and `unordered_map`'s  are currently used in Multigrid and `meshblock_tree.*` Can you say more about the maps / can the same maps be shared across these classes?

felker: @tomidakn did you see this question?

tomidakn: Sorry for my late response. I moved this as I just feel this operator should be associated with the LogicalLocation structure like other operators. The map is used to quickly find a neighboring MeshBlock without searching the whole Tree.

PR #636, opened 2024-11-19
Initialized turb_flag in the Mesh constructor

### LINKED ISSUE #634: Fix Turbulence Driver input flag ambiguity
The [documentation](https://github.com/PrincetonUniversity/athena/wiki/Turbulence-Driver) says that we need to set `turb_flag` in the `<problem>` block (and other turbulence parameters in the `<turbulence>` block). However, I found that this is not the case, and we need to set Mesh::turb_flag in a pgen file (typically in Mesh::InitMeshData), as in `turb.cpp`. This is inconsistent.

I would like to propose the following changes:
1. Move `turb_flag` to `<turbulence>`.
2. Read `turb_flag` automatically.

Or at very least, we should update the manual to be consistent with the current implementation.

changgoo:
## Prerequisite checklist

## Description

1. Initialized `turb_flag` in the Mesh constructor by reading it from the problem input under the `turbulence` block (with the default being 0). No need to initialize it in the pgen files. 

This closes #634

### DIFF
--- a/inputs/hydro/athinput.turb
+++ b/inputs/hydro/athinput.turb
@@ -63,6 +63,7 @@ gamma           = 1.666666666667 # gamma = C_p/C_v
 iso_sound_speed = 1.00           # equivalent to sqrt(gamma*p/d) for p=0.1, d=1
 
 <turbulence>
+turb_flag  = 1    # 1 for decaying, 2 (impulsive) or 3 (continuous) for driven turbulence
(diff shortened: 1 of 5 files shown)


felker (approved):

changgoo:
@felker wow, you merge it before I update wiki! Now, it is updated.

tomidakn:
Thank you very much!!

PR #638, opened 2024-12-05
Throw fatal runtime error for asymmetric x2 bounds for 1D spherical-polar mesh

### LINKED ISSUE #354: In spherical polar coordinates, theta limits not symmetric about pi/2 cause geometric source term errors
In spherical polar coordinates, if you set theta limits to not be symmetric about pi/2 then the geometric source terms give the wrong value. 

**Summary of issue**

This was only noticed in a 1D Parker wind calculation (i.e. nx2=nx3=1) and lead to an inaccurate solution.

PS: I'm on paternity leave, so I don't have time to fill this out properly, but still wanted to document it.

**Steps to reproduce**

Set x2min, x2max such that pi/2-x2min != x2max-pi/2.

apbailey:
## Prerequisite checklist

## Description
Add fatal error for 1d spherical polar problems if the x2 limits are not symmetric about PI/2. This addresses the raised issue #354. This also updates the x2-limits for the 1d mignone radial test to be symmetric so that the regression test doesn't conflict with this PR. 

## Testing and validation
tested against regression tests involving spherical polar coordinates

## To-do
does not address associated non-zero vtheta in 1d spherical in any way

Closes #354

### DIFF
--- a/inputs/hydro/athinput.mignone_radial
+++ b/inputs/hydro/athinput.mignone_radial
@@ -34,9 +34,9 @@ ix1_bc     = reflecting   # inner-X1 boundary flag
 ox1_bc     = outflow   # outer-X1 boundary flag
 x1rat      = 1.0       # non-uniform grid ratio
 
-nx2        =  1        # Number of zones in X2-direction
-x2min      = 0.0       # minimum value of X2
-x2max      = 1.0       # maximum value of X2
+nx2        =  1       # Number of zones in X2-direction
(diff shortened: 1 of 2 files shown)


buildbot-princeton:

felker:
add to whitelist

felker:
Good catch with the Mignone regression test. I guess the asymmetric limits did not cause any issues because of:
```
active    = background    # compute Hydro fluxes, etc. but do NOT evolve the fluid
```

felker (approved):

PR #650, opened 2025-03-07
Added new compiler configuraitons

tomidakn:
This PR adds new compiler options in the configure script.

Added:
 - AOCC (AMD version of clang++)
 - New CrayCC (based on clang++)

Removed:
 - bgxlc++ (IBM Blue Gene XLC++)

Planned:
 - Latest versions of icpx (-qopt-prefetch=4 crashes with v2024.0, 2024.2 is OK)

I will update the Wiki once this is approved.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -282,14 +282,15 @@ cxx_choices = [
     'g++',
     'g++-simd',
     'icpx',
+    'icpx-old',
     'icpc',
     'icpc-debug',
     'icpc-phi',
     'cray',
-    'bgxlc++',
     'clang++',
     'clang++-simd',
     'clang++-apple',
+    'aocc',
 ]
 
 
@@ -297,9 +298,6 @@ def c_to_cpp(arg):
     arg = arg.replace('gcc', 'g++', 1)
     arg = arg.replace('icc', 'icpc', 1)
     arg = arg.replace('icx', 'icpx', 1)
-    if arg == 'bgxl' or arg == 'bgxlc':
-        arg = 'bgxlc++'
-
     if arg == 'clang':
         arg = 'clang++'
     else:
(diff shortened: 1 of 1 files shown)


msbc:
Should we add nvc++ while we're at it? Nvidia is scheduled to release consumer grade ARM processors this year.

configure.py
```diff
@@ -649,40 +659,19 @@ def c_to_cpp(arg):
...
 if args['cxx'] == 'cray':
-    # Cray Compiling Environment 8.4 (2015-09-24) introduces C++11 feature completeness
-    # (except "alignas"). v8.6 is C++14 feature-complete
+    # New HPE Cray Compiling Environment based on clang (2019-)
     definitions['COMPILER_CHOICE'] = 'cray'
     definitions['COMPILER_COMMAND'] = makefile_options['COMPILER_COMMAND'] = 'CC'
     makefile_options['PREPROCESSOR_FLAGS'] = ''
-    makefile_options['COMPILER_FLAGS'] = '-O3 -h std=c++11 -h aggress -h vector3 -hfp3'
-    makefile_options['LINKER_FLAGS'] = '-hwp -hpl=obj/lib'
+    makefile_options['COMPILER_FLAGS'] = '-Ofast -std=c++11 -flto'
```
felker, line 666: what machine are you primarily targeting for the Cray compilers?  Do we really want `-Ofast` enabled by default?

tomidakn: National Astronomical Observatory of Japan recently introduced HPE Cray XD2000 with Sapphire Rapids. But honestly, there is not point to use Cray Compiler because they also have Intel oneAPI and it is faster by almost a factor of 2. So I think it is fine to simply remove it.  I just felt that the old one must be gone.

Regarding -Ofast, that was suggested on a manual of another computing center. I do not worry about that too much to be honest, but I am OK to change it to -O3. It seems that Clang does not recommend it anyway.
https://discourse.llvm.org/t/rfc-deprecate-ofast/78687

felker: Yeah I agree with the comment there:
> -ffast-math is a complete footgun, and having -Ofast include it only proliferates its use without proper understanding of the effects of it.

tomidakn: OK, I am happy to demote it to -O3.

PR #648, opened 2025-02-14
MHD reproducibility

tomidakn:
This commit improves reproducibility of parallel MHD simulations with a new EMF correction method, taking a maximum value among coinciding edges between MeshBlocks.  See issue #645.

## Prerequisite checklist
- [x] My code follows the Athena++ [Style Guide](https://github.com/PrincetonUniversity/athena/wiki/Style-Guide)
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation in the [Wiki](https://github.com/PrincetonUniversity/athena/wiki) accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

## Description
Previously we simply used average, but such operations are not reproducible because they depend on the ordering of arrival of MPI communications. This implementation is simpler than the previous one, as it does not need to count the number of coinciding edges.

Note 1: this improves the reproducibility between runs, but it does NOT reproduce calculations with different MeshBlock configuration (e.g. 1 MeshBlock vs multiple MeshBlocks).

Note 2: the operation can be minimum instead of maximum, or max/min in absolute values. I even tried switching max/min at every cycle, but I could not find any meaningful difference between them. All we need is to use exactly same EMF for coinciding edges. 

Note 3: I now feel this implementation is better than the previous one, but I'm not 100% sure yet.

This also contains minor leftover fixes.

PR #663, opened 2025-04-11
Revert "MHD reproducibility"

felker:
Reverts PrincetonUniversity/athena#648


tomidakn:
Thanks for taking care of this. Yes, this has to be reverted.

PR #667, opened 2025-04-28
Chemistry updates

munan:
## Prerequisite checklist

- [x ] My code follows the Athena++ [Style Guide](https://github.com/PrincetonUniversity/athena/wiki/Style-Guide)
- [x ] I have added tests to cover my changes.
- [x ] All new and existing tests passed.

## Description
Major changes:
- Make six-ray compatible with periodic boundary condition, modified regression test for that

Minor changes:
- Fixed chemical rates
- Add dependency of INT_CHM after INT_HYD

## Testing and validation
Re-run all chemistry regression tests and passed.


src/task_list/time_integrator.cpp
```diff
@@ -1026,7 +1026,7 @@ TimeIntegratorTaskList::TimeIntegratorTaskList(ParameterInput *pin, Mesh *pm) {
       } else {
         AddTask(INT_SCLR,CALC_SCLRFLX);
       }
-      AddTask(INT_CHM,INT_SCLR);
+      AddTask(INT_CHM,(INT_HYD|INT_SCLR));
```
felker, line 1029: I am having trouble following the graph of dependencies here. Was it a bug beforehand?

munan: This did not cause a bug so far. But when I was trying to figure out another bug and read through this, I realised that  the order of INT_HYD and INT_CHM may not be guaranteed if I do not add dependencies here. Since INT_CHM also update the (internal) energy, I think it is better to add the dependency. Let me know if it makes sense.

PR #673, opened 2025-05-21
Modification to the Units module

rjtanner:
## Prerequisite checklist

## Description

Major changes:
1. I reworked the units module so that it now explicitly works from a "unit basis" (something that was implicitly used in the original units module). The unit basis is used to convert from physical units to code units. The updated units module is backwards compatible with the previous version.
2. I modified the units module to make it easier to add additional units, and included a number of new basis units.

Minor changes:
1. I modified the module unit to include more preset unit systems.
2. Some minor changes to the names of a few constants to agree with the FITS standard for unit names.
3. I included a function that will output the units_override dictionary needed for reading in HDF5 files using yt.
4. The mean atomic weight can now be set in the input file.

## Testing and validation

This doesn't change anything "internal" to Athena, but only facilitates the interface between the input file and the problem file.

I included an example problem and input file that demonstrates how to use units.

## To-do

The wiki page on units would need to be updated. I can write something up for that.


buildbot-princeton:

tomidakn (commented):
It looks fine to me. Some variables names do not follow the style but I think Gconst and H(hydrogen) are acceptable. But I would like to leave the final decision to @changgoo .

I appreciate if you could write up the Wiki documentation and send it to me so that I can update it.

PR #678, opened 2025-06-21
Add runtime type setting for HDF5 outputs (via templates)

msbc:
## Prerequisite checklist

## Description

This PR makes the type (e.g., float vs double) of the HDF5 outputs a runtime option. Also, (will) add options for casting to unsigned int types with `vmin` and `vmax` setting how the floats get transformed into unsigned ints. Casting to an 8 bit uint is (i.e. 0-255) is useful if you only want the data to make plots. 256 is equal to the default number of colors used in a `matplotlib` colormap.

## Testing and validation

Um... See to-do.

...

## To-do

- [x] Add tests
- [x] Finish regression tests
- [x] Implement vmin/vmax logic


msbc:
This is definitely a work in progress.

changgoo:
@msbc You will receive a failure message from Jenkins including the log. The current failure is due to style:
```
Done processing ../../src/outputs/outputs.cpp
../../src/outputs/outputs.hpp:60:  Add #include <limits> for numeric_limits<>  [build/include_what_you_use] [4]
```

src/outputs/athena_hdf5.cpp
```diff
@@ -86,13 +86,13 @@ void ATHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, bool flag) {
...
   int *levels_mesh;                            // array of refinement levels on Mesh
   std::int64_t *locations_mesh;                // array of logical locations on Mesh
-  H5Real *x1f_mesh;                            // array of x1 values on Mesh
-  H5Real *x2f_mesh;                            // array of x2 values on Mesh
-  H5Real *x3f_mesh;                            // array of x3 values on Mesh
-  H5Real *x1v_mesh;                            // array of x1 values on Mesh
-  H5Real *x2v_mesh;                            // array of x2 values on Mesh
-  H5Real *x3v_mesh;                            // array of x3 values on Mesh
-  H5Real **data_buffers;                       // array of data buffers
+  T *x1f_mesh;                            // array of x1 values on Mesh
```
felker, line 89: might be better to rename this template type for readability?

msbc: Yes, `T` is probably not the most readable.  How about `out_t`? Compromise of short length and readability.

felker: `h5out_t`? I dont feel strongly

msbc:
If it's implemented as intended then hdf5 output files will default to fp32 without the `-h5double` flag and fp64 with that flag, to maintain backwards comparability. Also, I've also made some local changes/bug fixes that I still need to push (and commit).

felker:
is this ready to review and merge?

msbc:
Yes, I believe so. All that's left is to update the wiki.

PR #690, opened 2025-09-30
New reconstructions

tomidakn:
This adds a few new options for reconstructions.
- xorder=2m,2mc: 2nd-order PLM with the minmod limiter (for stability)
- xorder=3f: 3rd-order PPM without the extrema-preserving limiter and characteristic reconstruction (for performance)
- xorder=wenoz: 5th-order WENO-Z (Borges et al. 2006)
- xorder=wenomz: 5th-order WENO-MZ (Wang et al. 2024)

## Prerequisite checklist
- [X] My code follows the Athena++ [Style Guide](https://github.com/PrincetonUniversity/athena/wiki/Style-Guide)
- [X] My change requires a change to the documentation.

I added a test only for xorder=3f. 

For WENO-Z and WENO-ZM, the hydro4/hydro_linwave_2d test does not work well because these schemes converge very fast and reach saturation at very low resolution. I tested them and they are 5-th order up to Nx=64 or so.

### PLM with the minmod limiter
I often had to rewrite the limiter to improve the robustness when I am dealing with severe simulations. This change makes it official.

### PPM without the extrema-preserving limiter and characteristic reconstruction (PPM-Fast)
The current implementation of PPM is actually slow. This is mainly because of the characteristic reconstruction, as it has to be called for all the data points within the stencil. By omitting it, we can reuse some data and improve the performance.

This version follows the conventional PPM implementation of Colella & Woodward 1984, and does not support the extrema-preserving limiter. The extrema-preserving limiter is nice to improve the accuracy for idealized test problems (and is needed to achieve the formal accuracy), but often causes numerical problems.

### Option to turn off the extrema-preserving limiter in PPM
Also, now with the time/ext_preserving=false option, one can turn off the extrema-preserving limiter in PPM. But this is still slow, so I recommend to use xorder=3f unless you need the characteristic reconstruction.

### WENO-Z and WENO-MZ
I ported the WENO-Z (Borges et al. 2006) implementation from AthenaK. In addition, I implemented WENO-MZ (Wang et al. 2024), one of the latest implementations of WENO-Z type reconstructions. These require only 5-point stencil and the implementations are simple but can achieve higher accuracy.

## Testing and validation
I tested them using hydro linear wave tests. PLM with minmod is somewhere in between 1st and 2nd order accurate. PPM-Fast and PPM without the extrema-preserving limiter are 2nd-order (because of the extrema clipping). WENO-Z and WENO-MZ are 5th order.

## To-do
- Add regression tests to cover all the cases.
- Implement a higher-order CT scheme for MHD


jmstone (approved):
Adding additional reconstruction options is useful.  I may steal the WENOMZ to use in AthenaK!

felker:
if I recall correctly, the code already has something like `3f` in it? If the mesh is nonuniform and/or curvilinear, we use Mignone (2014) which clips smooth extrema?
https://github.com/PrincetonUniversity/athena/blob/e39297859810e5e95b45b59e9004bdf8b8c5a5a3/src/reconstruct/ppm_simple.cpp#L171-L174

how does your implementation differ?

tomidakn:
> if I recall correctly, the code already has something like `3f` in it? If the mesh is nonuniform and/or curvilinear, we use Mignone (2014) which clips smooth extrema?
> how does your implementation differ?

In my implementation, if you specify time/xorder=3 and time/ext_preserving=false, it turns off the extrema preserving limiter and uses the Mignone limiter. However, this implementation is significantly slower than the best implementation of PPM. This is because the code does not reuse the reconstructed data from the previous cycle even when it is possible, and this is due to the characteristic reconstrction. PPM-Fast (time/xorder=3f) omits the characteristic reconstruction and reuses the data as much as possible. 

So, in terms of the accuracy, PPM + ext_preserving=false is very similar to PPM-Fast (at least on Cartesian, uniformly-spaced grids). The main difference is the performance. I have measured the performance with the 2D Hydro linear wave test using RK3 with 1 core (Nx1=128, Nx2=64),
PPM with the extrema preserving limiter: 1.30e6 cells/sec
PPM without the extrema preserving limiter: 1.45e6 cells/sec
PPM-Fast: 2.52e6 cells/sec
PLM: 3.30e6 cells/sec

tomidakn:
@ajdittmann Because WENO-Z+M (and WENO-Z+) explicitly depends on delta x, which means the scheme is not scale independent.

ajdittmann:
> Because WENO-Z+M (and WENO-Z+) explicitly depends on delta x, which means the scheme is not scale independent.
WENO-Z+ does, but WENO-Z+M (as in Luo and Wu 2021) sets lambda to a constant, rater than a function of delta x. 
I ask specifically because a comment in your earlier commits mentioned that Z+M preserved symmetry better than MZ.

tomidakn:
Ah, I forgot that because Wang et al. 2024 (WENO-MZ paper) argued that WENO-Z+M is scale dependent, but you are right. In that sense, however, we have to choose the right value of lambda, and it looks like it is determined empirically.

tomidakn:
But honestly, I am not saying that WENO-MZ is the best WENO variant. I just wanted to try something better than WENO-Z+, and chose WENO-MZ because it looked theoretically simpler. We can add more variant if there is anything better.

PR #720, opened 2026-05-21
fix(amr): correct sffc3 buffer size for non-cubic meshblocks (#719)

### LINKED ISSUE #719: Wrong Size Allocation in PrepareAndSendFaceFieldCorrection
### Prerequisite Checklist

- [x] I have checked that this issue hasn't already been reported
- [x] I have checked the documentation in the Wiki for an answer
- [x] I am running the latest version of Athena++

### Summary of Issue

In line 1231 of amr_loadbalance.cpp, the sizes of field buffers are calculated as

` const int sffc1 = nf*bnx2*bnx3, sffc2 = nf*bnx1*bnx3, sffc3 = nf*bnx2*bnx3;`

The last variable should be 

`sffc3 = nf*bnx1*bnx2'

Otherwise a segfault is produced if the mesh doesn't have the same nx1, nx2, and nx3.  

### Steps to Reproduce

Run AMR with magnetic fields with nx1=nx2 and nx3!=nx1

### Expected Outcome

_No response_

### Actual Outcome

_No response_

### Additional Comments and/or Proposed Solution

_No response_

### Athena++ Version

_No response_

### Compiler and Version

_No response_

### Operating System

_No response_

### Hardware and Cluster Name

_No response_

### External Library Versions

_No response_

SAY-5:
## Description

Fixes #719.

In `Mesh::PrepareAndSendFaceFieldCorrection` (`src/mesh/amr_loadbalance.cpp:1231`), the per-face buffer sizes `sffc1`, `sffc2`, `sffc3` are set to `nf*bnx2*bnx3`, `nf*bnx1*bnx3`, and `nf*bnx2*bnx3` respectively. The third one is wrong: x3-face buffers carry the XY-plane (`bnx1*bnx2`), not the YZ-plane (`bnx2*bnx3`).

The downstream pack call confirms it:

```cpp
case BoundaryFace::inner_x3:
  BufferUtility::PackData(var_fc.x3f, f.buf, pmb->is, pmb->ie,
                          pmb->js, pmb->je, pmb->ks, pmb->ks, p);
```

It writes `(ie-is+1) * (je-js+1) * 1 = bnx1 * bnx2` doubles per field, so the buffer must be `nf*bnx1*bnx2`. With the current expression, whenever `bnx1 != bnx3` the x3-face buffer is wrong-sized: either too small (out-of-bounds write/MPI send) or too large (silently sends garbage). As reported, this produces a segfault for meshes with unequal block dimensions.

## Changes

```diff
-  const int sffc1 = nf*bnx2*bnx3, sffc2 = nf*bnx1*bnx3, sffc3 = nf*bnx2*bnx3;
+  const int sffc1 = nf*bnx2*bnx3, sffc2 = nf*bnx1*bnx3, sffc3 = nf*bnx1*bnx2;
```

## Testing and validation

- `tst/regression/run_tests.py amr/amr_linwave` passes both before and after the change.
- Verified the diff manually against the matching pack/unpack call shapes in the same function for inner_x3 and outer_x3.

## To-do

- A dedicated regression test that exercises `PrepareAndSendFaceFieldCorrection` with `bnx1 != bnx3` (3D MHD AMR with non-cubic meshblocks) would catch this class of bug; happy to follow up if useful.

### DIFF
--- a/src/mesh/amr_loadbalance.cpp
+++ b/src/mesh/amr_loadbalance.cpp
@@ -1228,7 +1228,7 @@ void Mesh::PrepareAndSendFaceFieldCorrection(LogicalLocation *newloc,
   const int bnx2 = my_blocks(0)->block_size.nx2;
   const int bnx3 = my_blocks(0)->block_size.nx3;
   const int nf = static_cast<int>(my_blocks(0)->pmr->pvars_fc_.size());
-  const int sffc1 = nf*bnx2*bnx3, sffc2 = nf*bnx1*bnx3, sffc3 = nf*bnx2*bnx3;
+  const int sffc1 = nf*bnx2*bnx3, sffc2 = nf*bnx1*bnx3, sffc3 = nf*bnx1*bnx2;
 
   // Step FFC1. Construct a hash map of MeshBlocks before refinement
   for (int i = 0; i < nbtold; ++i) {


tomidakn (approved):
This is an embarrassing bug, and I appreciate you have fixed it.

